PR #38, opened 2021-04-23
Added restart, reset disk and shutdown functions to the command line

eric-c-wood:


src/go/cmd/vm.go
```diff
@@ -123,6 +123,68 @@ func newVMResumeCmd() *cobra.Command {
...
+}
+
+func newVMResetDiskCmd() *cobra.Command {
+	desc := `Resets the disk state to the initial pre-boot disk state for a running or powered off VM
+	
+  Used to reset the disk state for the first disk for a running or powered off virtual machine for a specific 
+  experiment.  The VM's snapshot flag must be set to true in order to use this command.`
+
+	cmd := &cobra.Command{
+		Use:   "resetdisk <experiment name> <vm name>",
```
activeshadow, line 161: @eric-c-wood can we hyphenate this subcommand name to keep things consistent?

activeshadow (changes requested):

src/go/cmd/vm.go
eric-c-wood: @activeshadow hyphen has been added.

activeshadow (approved):

PR #41, opened 2021-04-27
Extended VM search fields

eric-c-wood:
The VM search fields were extended to allow additional queries/searches.

Here are some example searches:

1. 192.168.1.0/24 and dnb (Stopped experiment view to find VMs with interfaces in the specified subnet and have the do not boot flag set to true)
2. dnb (Stopped experiment view to find all VMs with the do not boot flag set to true)
3. capturing and router (Running experiment view to find all VMs that have an active packet capture and have router appearing in the name, host, taps, tags, or disk fields)
4. shutdown and 10.0.0.0/8 (Running experiment view to find all VMs that have been shutdown and have interfaces in the specified subnet)
5. 192.168.1.0/24 and 10.0.0.0/8 (Running/Stopped experiment views to find all VMs with interfaces in both the specified subnets)

The debounce function from the lodash library was used to wait some period of time before sending the search query to the server. https://lodash.com/docs/#debounce


glattercj:
I had a quick thought and tested it out: what happens if a VM is named one of the keywords? E.g. a VM named `dnb`? Result: The VMs with `do not boot` flag set to true show up in search results, but the VM named `dnb` does not. I'm assuming it's the same with the other keywords. Should we prefix or surround the keywords somehow with a special character perhaps? Like `.`? @activeshadow thoughts?

eric-c-wood:
@glattercj , thanks for taking a look at this.  Unfortunately, as you discovered if a keyword is the VM name, the VM name and the other default search fields will not be searched for the keyword.  Let me know what keyword escape character(s) you would like to use.  As a quick patch/test, I ended up surrounding a keyword with single or double quotes to avoid treating the term as a keyword.  I was going to use other well known wildcard characters such as `*%` but I ran into some conflicts that I would need to take more time to properly understand.

On another note, after looking at the regular expressions, if a search term contained any keyword, the entire term was incorrectly treated as a keyword.  For example, if there was a VM named free_running and the search term was `_running`, the state field would have been incorrectly searched for the state `_running` which would have yielded zero results.  Based on the existing patch, the free_running VM can be found with the search term `"running"`.  To search for the running state, the keyword `running` without quotes can be entered.

activeshadow:
@eric-c-wood quoting keywords to avoid treating them as keywords seems like a good approach, but then that precludes one from searching for keyword terms on VMs with a name that is the keyword, correct? For example, if you have a VM named `dnb`, then you'll not be able to search to see if `dnb` is marked as do not boot, correct? Not that one would ever want to do this, but I'm just trying to understand the implications.

Couple of options... prevent VMs from being named the same as keywords; use qualifiers when searching (e.g. `dnb:true` or `ip:192.168.1.0/24`). I like the second approach, but the first approach might be easier to implement. Or... leaving it as you have it now (quoting keywords) is fine too.

Just my two cents.

eric-c-wood:
@activeshadow Search terms can be combined.  If I understood your example correctly, then you can search for a VM that has `dnb` in the name by using quotes `"dnb"` but you can also search for the do not boot flag by using `dnb` without quotes.  I am posting a few screenshots as an example.

In the screenshot below, there are 2 VMs with the do not boot flag set to true.  There are also 2 VMs that have `dnb` as part of the VM name.
VMs_with_dnb_donotboot


To find the VMs with `dnb` as part of the VM name, the following search could be executed.
dnb_name

Now to find the VM with `dnb` as the VM name and has the do not boot flag set to true, the following search could be executed.
dnb_name_and_do_not_boot

Of course, if you want to find the VM with `dnb` as part of the VM name but had the do not boot flag set to false, you could execute the following search
dnb_name_and_not_do_not_boot

Let me know if I failed to understand your example and address your concerns. 

With regards to using qualifiers where the user essentially defines the fields to search, we had implemented something like that previously.  While using qualifiers works and is easier to implement, for a lazy user like myself, I found it cumbersome especially as the number of search fields becomes large.  On the other hand, using qualifiers allows the user to be very clear in the intent of the search especially when a term can be found in multiple fields.  Perhaps a hybrid solution would be appropriate where qualifiers would be optional.

activeshadow:
@eric-c-wood cool, I'll just merge this in as-is then. Looks like it's working fine from the screenshots.

PR #46, opened 2021-05-14
Added checkboxes and a multi-selection boot flag to stopped experiment view

eric-c-wood:
Added checkboxes to the stopped experiment view to enable future actions to selected VMs.  As an example use case, a multi-selection boot flag was added.

**Multi-Selection Boot Flag Examples**
Select some VMs and click the lightning bolt under the `Boot` label.
bulk_boot_flag

The search bar can also be used to create the selection set
bulk_boot_flag_selection_set


glattercj (approved):

activeshadow:
@eric-c-wood you are a machine! I do echo @glattercj's comment about grouping toasts. Thoughts?

eric-c-wood:
@glattercj @activeshadow With regards to grouping the toasts, is there a preference for one toast that list all the VMs that had the `dnb` flag updated? (e.g. VM1, VM2, VM3, were set to "boot" or "do not boot")  Alternatively, would a generic toast that does not list all the VMs be acceptable? (e.g. The selected VMs were set to "boot" or "do not boot")  After all, there is also the lightning bolt that gives a visual indication of the boot flag.  The toast is slightly redundant feedback.

activeshadow:
@eric-c-wood I think a generic toast would be better.

glattercj:
👍  generic toast. FYI I think the latest merge created conflicts on a couple of your PRs.

glattercj:
@eric-c-wood Ready to merge after conflicts are resolved.

PR #61, opened 2021-06-12
Enable using minimega C2 on Windows 10 VMs to run startup script

activeshadow:
This obviously requires that the Windows 10 VM image be configured ahead
of time to run the miniccc agent at boot. Since the `image inject-miniexe`
subcommand uses the startup folder script technique to start the miniccc
agent, the miniccc agent and service must be added to Windows 10 images
in a different manner (e.g. a Packer config).

### DIFF
--- a/src/go/app/startup.go
+++ b/src/go/app/startup.go
@@ -5,8 +5,10 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"phenix/internal/common"
+	"phenix/internal/mm"
 	"phenix/tmpl"
 	"phenix/types"
 	ifaces "phenix/types/interfaces"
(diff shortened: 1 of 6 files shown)


glattercj (approved):

glattercj:
@activeshadow Might want to put in the subcommand help where the file is expected to be i.e. `/startup.ps1`.

activeshadow:
> @activeshadow Might want to put in the subcommand help where the file is expected to be i.e. `/startup.ps1`.

@glattercj the `startup` app still takes care of where the `startup.ps1` script is placed in the VM, and the path that miniccc is instructed to execute is not customizable by the user, so I'm wondering why you think the documentation needs to be updated?

glattercj:
> > @activeshadow Might want to put in the subcommand help where the file is expected to be i.e. `/startup.ps1`.
> 
> @glattercj the `startup` app still takes care of where the `startup.ps1` script is placed in the VM, and the path that miniccc is instructed to execute is not customizable by the user, so I'm wondering why you think the documentation needs to be updated?

I guess I misunderstood. I was meaning if this command's success depends on certain files being somewhere a priori then those should be included with this command's docs. If that's already done, ignore me :)

PR #62, opened 2021-06-14
Update Dockerfile

glattercj:
Add optional ROOT_CERT for docker build.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -25,11 +25,14 @@ WORKDIR /phenix
 
 COPY --from=jsbuilder /phenix/src/js /phenix/src/js
 
-ARG INSTALL_CERT=
-RUN if [ -n "${INSTALL_CERT}" ]; then \
-  wget ${INSTALL_CERT} -e use_proxy=no \
-    -O /usr/local/share/ca-certificates/custom.crt \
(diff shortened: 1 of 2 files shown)


docker/Dockerfile
```diff
@@ -31,6 +31,12 @@ RUN if [ -n "${INSTALL_CERT}" ]; then \
     -O /usr/local/share/ca-certificates/custom.crt \
     && update-ca-certificates; fi
 
+ARG ROOT_CERT=
```
activeshadow, line 34: @glattercj I'm wondering if we shouldn't just check for a comma-separated list of cert URLs for `INSTALL_CERT`? That would make it future-proof. Otherwise, can we use `custom-root.crt` instead of `root.crt` for this new certificate?

activeshadow (changes requested):

docker/Dockerfile
glattercj: Yes, definitely. I was just being lazy...

activeshadow: Sheesh... like you have something better to be doing. :joy:

glattercj: If you can make it prettier please do.

activeshadow (approved):
LGTM

PR #63, opened 2021-06-16
Support minirouter as router os_type

activeshadow:
When minirouter is specified as the os_type for nodes of type Router,
the minimega router commands are used to configure the VM instead of
injecting config files.

Currently supports configuring interface addresses, default gateways,
static routes, and OSPF routing. At some point it will be updated to
also support ACLs via iptables configurations.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,8 +1,7 @@
+*.bdb
+*.code-workspace
+
 .devcontainer
 .vscode
-*.bdb
 
 bin
-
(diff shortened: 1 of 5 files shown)


activeshadow:
@glattercj take another look when you get a chance. I've updated my local topology I'm testing with to use the `proto: ospf` interface setting for router interfaces and am testing now.

glattercj (approved):

PR #80, opened 2021-10-22
Add banner to VM VNC windows

activeshadow:
If a VM has a label named `vncBanner` in its topology config, then the
value of the label will be shown at the top and bottom of the VNC window
in yellow letters inside of a div with a red background.

### DIFF
--- a/src/go/api/vm/vm.go
+++ b/src/go/api/vm/vm.go
@@ -160,16 +160,18 @@ func Get(expName, vmName string) (*mm.VM, error) {
 		}
 
 		vm = &mm.VM{
-			ID:         idx,
-			Name:       node.General().Hostname(),
-			Experiment: exp.Spec.ExperimentName(),
-			CPUs:       node.Hardware().VCPU(),
-			RAM:        node.Hardware().Memory(),
-			Disk:       node.Hardware().Drives()[0].Image(),
-			Interfaces: make(map[string]string),
-			DoNotBoot:  *node.General().DoNotBoot(),
-			OSType:     string(node.Hardware().OSType()),
-			Metadata:   make(map[string]interface{}),
+			ID:          idx,
+			Name:        node.General().Hostname(),
+			Experiment:  exp.Spec.ExperimentName(),
+			CPUs:        node.Hardware().VCPU(),
+			RAM:         node.Hardware().Memory(),
+			Disk:        node.Hardware().Drives()[0].Image(),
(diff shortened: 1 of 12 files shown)


glattercj:
`s/topology/scenario`?

glattercj:
Nvm, I guess you can add `labels:` to both topology and scenario.

activeshadow:
@aherna please take a look at the latest version of this PR when you have time. Please note the following changes when doing so:

* VNC banner configuration is now a node annotation (as opposed to a node label).
* The value of the annotation can either be a string or a map.
  * If it's a string, the value is assumed to be the label to use for the top and bottom banner (as a single line - no multi-line banner support), and a white background with black text will be used for the banners.
  * If it's a map, it's expected to have the following schema:

```
annotations:
  vncBanner:
    topBanner:
      banner:
        - Top
        - Banner
      backgroundColor: red
      textColor: yellow
    bottomBanner:
      banner:
        - Bottom Banner
      backgroundColor: green
      textColor: red
```
Both `topBanner` and `bottomBanner` sections are optional. The `banner` value is an array of strings, each being its own line in the banner (thus enabling multi-line banners). Values for `backgroundColor` and `textColor` can be anything CSS recognizes as color values (to include HTML color codes).

glattercj:
@activeshadow 
Tested using the example above, plus a single vncBanner string on a different node:

Firefox: Both top and bottom banner show as expected
Chrome: Only top banner shows
Safari: Both top and bottom banner show as expected

activeshadow:
> Chrome: Only top banner shows

Ugh...

activeshadow:
@aherna @glattercj do you want this working for Chrome as a target browser, or is Firefox and Safari support sufficient?

aherna:
> @activeshadow Chrome and Firefox are the primary browsers. We can live with firefox for now and make hotfix later for chrome? @glattercj is that good with you?

glattercj:
> > @activeshadow Chrome and Firefox are the primary browsers. We can live with firefox for now and make hotfix later for chrome? @glattercj is that good with you?

Sounds good, unless it's a quick hotfix then might as well add it to this PR...

activeshadow:
> Sounds good, unless it's a quick hotfix then might as well add it to this PR...

@glattercj I have no idea if it will be quick or not... cross-browser styling support is not my strong suit. I'll do a little poking and if I don't find anything I'll go ahead and merge.

PR #84, opened 2021-11-19
ID cc clients using UUID by default instead of topology hostname

activeshadow:
Added an option to ID clients by hostname, defaulting to false, which
means by default clients will be identified by their UUID when checking
to see if C2 is active for a client.

### DIFF
--- a/src/go/api/soh/app.go
+++ b/src/go/api/soh/app.go
@@ -383,6 +383,10 @@ func (this *SOH) getFlows(ctx context.Context, exp *types.Experiment) {
 
 		opts := []mm.C2Option{mm.C2NS(exp.Metadata.Name), mm.C2VM(hostname), mm.C2Command("query-flows.sh")}
 
+		if this.md.useUUIDForC2Active(hostname) {
+			opts = append(opts, mm.C2IDClientsByUUID())
+		}
+
 		id, err = mm.ExecC2Command(opts...)
 		if err != nil {
 			if errors.Is(err, mm.ErrC2ClientNotActive) {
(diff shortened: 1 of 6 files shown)


activeshadow:
I don't want to merge this until the previous default mode of using the hostname to determine if C2 is active is possible again.

This is mainly used for the state of health app, so I propose adding a key like `useUUIDForC2Active` to the SoH config metadata, whose value would be a list of topology nodes that the SoH app should use the VM UUID for when determining if C2 is active instead of the VM hostname.

activeshadow:
@aherna please test this branch at your earliest convenience. Below is an experiment you can use to test with if you like.

```
apiVersion: phenix.sandia.gov/v1
kind: Experiment
metadata:
  name: cc-uuid-test
spec:
  scenario:
  - name: soh
    metadata:
      c2Timeout: 1m
      exitOnError: true
      skipInitialNetworkConfigTests: false
      testReachability: off
#     hostsToUseUUIDForC2Active:
#     - VM1
  topology:
    nodes:
    - general:
        hostname: VM1
        snapshot: false
      hardware:
        drives:
        - image: miniccc.qc2
        memory: 512
        vcpus: 1
      network:
        interfaces:
        - address: 192.168.12.10
          mask: 24
          name: IF0
          proto: static
          type: ethernet
          vlan: EXP
      type: VirtualMachine
```
Given the above experiment configuration, the VM's hostname will not be set to the hostname specified in the topology config since snapshot is disabled for the VM. The default for the state of health app is to use the hostname to determine if the cc client is active, so with this configuration the state of health app should fail after 1 minute and the experiment will be stopped.

Once confirming that the experiment does fail, you can uncomment the two commented lines in the experiment config above to use the VM's UUID when determining of the cc client is active and the state of health app should succeed.

PR #88, opened 2022-02-24
Update Windows startup process to add a registry entry when done

activeshadow:
The following is an example of a function other scripts can use to know
if the phenix startup script has fully completed.

```
function Phenix-StartupComplete() {
    $key = Get-Item -LiteralPath 'HKLM:\Software\phenix' -ErrorAction SilentlyContinue

    if ($key) {
        $val = $key.GetValue('startup')

        if ($val) {
          return $val -eq 'done'
        }

        return $false
    }

    return $false
}
```

This commit also updates the scheduled task to run a script that will
execute all PowerShell scripts present in `C:\phenix\startup\*.ps1` on
each boot. This means any scripts placed in this directory must be
idempotent or self-delete.

### DIFF
--- a/src/go/app/ntp.go
+++ b/src/go/app/ntp.go
@@ -92,7 +92,7 @@ func (NTP) PreStart(ctx context.Context, exp *types.Experiment) error {
 				return fmt.Errorf("generating Windows NTP script: %w", err)
 			}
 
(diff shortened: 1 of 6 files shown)


glattercj (approved):
LGTM

PR #93, opened 2022-04-23
Update vmdb template

glattercj:
Add `mount-virtual-filesystems` to support running services (e.g., docker) inside chroot during build. Moved fstab and grub earlier to avoid conflict with virtual filesystem mounts.


activeshadow:
@glattercj with at least one Ubuntu-based image config, this change is preventing image build scripts from being able to resolve domain names. After further investigation, `/etc/resolv.conf` is no longer present and trying to add it as part of one of the build scripts suggests the entire `/etc` directory is missing.

Are you seeing similar behavior?

PR #95, opened 2022-04-28
Ubuntu18.04.6:Files tab date column populated

eric-c-wood:
This fixes the **[UI] Ubuntu 18.04.6 Files Tab Date Column Empty** issue.  The problem was a hardcoded absolute path to the `ls` command.  Since the location of the `ls` command can vary by Linux distribution, the absolute path reference was removed.  The minimega shell command will call `exec.LookPath` to find the absolute path to a command as long as the command is listed in the PATH environment variable.   The screenshot below shows that the issue has been fixed.
date_column_18 04_populated


glattercj (approved):
LGTM

PR #99, opened 2022-05-23
Update ntp_windows script to work with Win10

cmulk:
Windows 10 has a slightly different output format when calling the `w32tm /monitor` command. This update makes the windows ntp script more flexible in getting the current `Stratum` value so it works with Win7 or Win10


activeshadow:
Thanks @cmulk!

PR #109, opened 2022-11-03
Add Ability to Mount VMs and Download/Upload Files

jacdavi:
When interacting with a VM in a running experiment, users can now access
files within the VM directly from the UI by clicking the `mount vm`
button within the VM's informational modal.

RBAC policies are applied to limit the ability for users to mount and
unmount VMs, as well as list, download, and upload files to and from
VMs.

Note that this feature only works if the miniccc agent is active within
the VM. If it's not, the `mount vm` button will be disabled.

Also note that this feature is also currently behind a feature flag in
the phenix UI named `vm-mount` that is not enabled by default. To enable
it, pass the `--features=vm-mount` flag to the `phenix ui` command when
starting the UI process.

Additional notes and changes:

* buefy dependency was upgraded to 0.9.22
  * this enabled use of the "breadcrumbs" component in the mount modal
  * this also required updating all tables due to a breaking change
* added flag for "cc active" to the frontend which is visible in the vm
  info modal
* created common vue filter for formatting file sizes
* fixed developer jwt signing key and added comment documenting use
* generalized ExperimentFile object to File and removed some unused code
  for ExperimentFiles in protobuf
* changed ESLint rules to not warn of unused variables if the variable
  starts with _

### DIFF
--- a/src/go/util/file/types.go
+++ b/src/go/util/file/types.go
@@ -1,6 +1,8 @@
 package file
 
 import (
+	"io/fs"
+	"path/filepath"
 	"sort"
 	"strings"
 	"time"
@@ -22,21 +24,22 @@ type ImageDetails struct {
 	Size     int
 }
 
-type ExperimentFile struct {
+type File struct {
 	Name       string   `json:"name"`
 	Path       string   `json:"path"`
 	Date       string   `json:"date"`
-	Size       int      `json:"size"`
+	Size       int64    `json:"size"`
 	Categories []string `json:"categories"`
 	PlainText  bool     `json:"plainText"`
+	IsDir	   bool     `json:"isDir"`
 
 	// Internal use to aid in sorting
 	dateTime time.Time
 }
 
-type ExperimentFiles []ExperimentFile
+type Files []File
 
-func (this ExperimentFiles) SortByName(asc bool) {
+func (this Files) SortByName(asc bool) {
 	sort.Slice(this, func(i, j int) bool {
 		if asc {
 			return strings.ToLower(this[i].Name) < strings.ToLower(this[j].Name)
@@ -46,7 +49,7 @@ func (this ExperimentFiles) SortByName(asc bool) {
 	})
 }
 
-func (this ExperimentFiles) SortByDate(asc bool) {
+func (this Files) SortByDate(asc bool) {
 	sort.Slice(this, func(i, j int) bool {
 		if asc {
 			return this[i].dateTime.Before(this[j].dateTime)
@@ -56,7 +59,7 @@ func (this ExperimentFiles) SortByDate(asc bool) {
 	})
 }
 
-func (this ExperimentFiles) SortBySize(asc bool) {
+func (this Files) SortBySize(asc bool) {
 	sort.Slice(this, func(i, j int) bool {
 		if asc {
 			return this[i].Size < this[j].Size
@@ -67,7 +70,7 @@ func (this ExperimentFiles) SortBySize(asc bool) {
 }
 
 /*
-func (this ExperimentFiles) SortByCategory(asc bool) {
+func (this Files) SortByCategory(asc bool) {
 	sort.Slice(this, func(i, j int) bool {
 		if asc {
 			return strings.ToLower(this[i].Category) < strings.ToLower(this[j].Category)
@@ -78,7 +81,7 @@ func (this ExperimentFiles) SortByCategory(asc bool) {
 }
 */
 
-func (this ExperimentFiles) SortBy(col string, asc bool) {
+func (this Files) SortBy(col string, asc bool) {
 	switch col {
 	case "name":
 		this.SortByName(asc)
@@ -91,14 +94,14 @@ func (this ExperimentFiles) SortBy(col string, asc bool) {
 	}
 }
 
-func (this ExperimentFiles) Paginate(page, size int) ExperimentFiles {
+func (this Files) Paginate(page, size int) Files {
 	var (
 		start = (page - 1) * size
 		end   = start + size
 	)
 
 	if start >= len(this) {
-		return ExperimentFiles{}
+		return Files{}
 	}
 
 	if end > len(this) {
@@ -106,4 +109,17 @@ func (this ExperimentFiles) Paginate(page, size int) ExperimentFiles {
 	}
 
 	return this[start:end]
+}
+
+// create an instance of our file object using a FileInfo and path
+func MakeFile(file fs.FileInfo, basePath string) (File) {
+	f := File{}
+	f.Name = file.Name()
+	f.Path = filepath.Join(basePath, file.Name())
+	f.dateTime = file.ModTime()
+	f.Date = file.ModTime().Format(time.RFC3339)
+	f.Size = file.Size()
+	f.IsDir = file.IsDir()
+
+	return f
 }
--- a/src/go/util/mm/minimega.go
+++ b/src/go/util/mm/minimega.go
@@ -4,7 +4,9 @@ import (
 	"bufio"
 	"context"
 	"encoding/base64"
+	"errors"
 	"fmt"
+	"os"
 	"path/filepath"
 	"regexp"
 	"strconv"
@@ -132,6 +134,9 @@ func (Minimega) GetLaunchProgress(ns string, expected int) (float64, error) {
 func (this Minimega) GetVMInfo(opts ...Option) VMs {
 	o := NewOptions(opts...)
 
+	// don't rely on `cc_active` column in `vm info` table
+	activeC2 := getActiveC2(o.ns)
+
 	cmd := mmcli.NewNamespacedCommand(o.ns)
 	cmd.Command = "vm info"
 	cmd.Columns = []string{"uuid", "host", "name", "state", "uptime", "vlan", "tap", "ip", "memory", "vcpus", "disks", "snapshot", "tags"}
@@ -143,14 +148,14 @@ func (this Minimega) GetVMInfo(opts ...Option) VMs {
 	var vms VMs
 
 	for _, row := range mmcli.RunTabular(cmd) {
-		var vm VM
-
-		vm.UUID = row["uuid"]
-		vm.Host = row["host"]
-		vm.Name = row["name"]
-
-		vm.Running = row["state"] == "RUNNING"
-		vm.State = row["state"]
+		vm := VM{
+			UUID:     row["uuid"],
+			Host:     row["host"],
+			Name:     row["name"],
+			State:    row["state"],
+			Running:  row["state"] == "RUNNING",
+			CCActive: activeC2[row["uuid"]],
+		}
 
 		s := row["vlan"]
 		s = strings.TrimPrefix(s, "[")
@@ -712,7 +717,6 @@ func (Minimega) GetVLANs(opts ...Option) (map[string]int, error) {
 
 func (Minimega) IsC2ClientActive(opts ...C2Option) error {
 	o := NewC2Options(opts...)
-
 	if o.skipActiveClientCheck {
 		return nil
 	}
@@ -766,6 +770,8 @@ func (this Minimega) ExecC2Command(opts ...C2Option) (string, error) {
 		return "", fmt.Errorf("cannot execute command: %w", err)
 	}
 
+	o := NewC2Options(opts...)
+
 	exec := func(ns, vm, cmd string) (string, error) {
 		ccMu.Lock()
 		defer ccMu.Unlock()
@@ -778,16 +784,19 @@ func (this Minimega) ExecC2Command(opts ...C2Option) (string, error) {
 		}
 
 		c.Command = cmd
+		c.Timeout = o.timeout
 
 		data, err := mmcli.SingleDataResponse(mmcli.Run(c))
 		if err != nil {
+			if errors.Is(err, mmcli.ErrTimeout) {
+				return "", fmt.Errorf("timeout running '%s' in vm %s", cmd, vm)
+			}
+
 			return "", fmt.Errorf("running '%s' in vm %s: %w", cmd, vm, err)
 		}
 
 		return fmt.Sprintf("%v", data), nil
 	}
-
-	o := NewC2Options(opts...)
 
 	if o.testConn != "" {
 		cmd := fmt.Sprintf("cc test-conn %s", o.testConn)
@@ -843,6 +852,33 @@ func (this Minimega) ExecC2Command(opts ...C2Option) (string, error) {
 		}
 
 		return id, nil
+	}
+
+	if o.mount != nil {
+		if *o.mount {
+			var (
+				path = GetLocalMountPath(o.ns, o.vm)
+				cmd  = fmt.Sprintf("cc mount %s %s", o.vm, path)
+			)
+
+			os.MkdirAll(path, os.ModePerm)
+
+			id, err := exec(o.ns, o.vm, cmd)
+			if err != nil {
+				return "", fmt.Errorf("Error creating mount: %w", err)
+			}
+
+			return id, nil
+		} else {
+			cmd := fmt.Sprintf("clear cc mount %s", o.vm)
+
+			id, err := exec(o.ns, o.vm, cmd)
+			if err != nil {
+				return "", fmt.Errorf("Error clearing mount: %w", err)
+			}
+
+			return id, nil
+		}
 	}
 
 	return "", fmt.Errorf("no options to execute were provided")
@@ -1055,6 +1091,25 @@ func (Minimega) MeshSend(ns, host, command string) error {
 	}
 
 	return nil
+}
+
+// GetLocalMountPath returns where the mount path should be on this filesystem
+// for the given namespace and VM.
+func GetLocalMountPath(ns string, vm string) string {
+	return filepath.Join(common.PhenixBase, "mounts", ns, vm)
+}
+
+func getActiveC2(ns string) map[string]bool {
+	active := make(map[string]bool)
+
+	cmd := mmcli.NewNamespacedCommand(ns)
+	cmd.Command = "cc client"
+
+	for _, row := range mmcli.RunTabular(cmd) {
+		active[row["uuid"]] = true
+	}
+
+	return active
 }
 
 func waitForResponse(ctx context.Context, ns, id string, timeout time.Duration) error {
--- a/src/go/web/server.go
+++ b/src/go/web/server.go
@@ -70,6 +70,7 @@ func Start(opts ...ServerOption) error {
 		}
 	}
 
+	router.HandleFunc("/features", GetFeatures).Methods("GET")
 	router.HandleFunc("/version", GetVersion).Methods("GET")
 	router.HandleFunc("/builder", GetBuilder).Methods("GET")
 	router.HandleFunc("/builder/save", SaveBuilderTopology).Methods("POST")
@@ -170,6 +171,15 @@ func Start(opts ...ServerOption) error {
 	api.HandleFunc("/experiments/{exp}/vms/{name}/snapshots/{snapshot}", RestoreVM).Methods("POST", "OPTIONS")
 	api.HandleFunc("/experiments/{exp}/vms/{name}/commit", CommitVM).Methods("POST", "OPTIONS")
 	api.HandleFunc("/experiments/{exp}/vms/{name}/memorySnapshot", CreateVMMemorySnapshot).Methods("POST", "OPTIONS")
+
+	if o.featured("vm-mount") {
+		api.HandleFunc("/experiments/{exp}/vms/{name}/mount", MountVM).Methods("POST", "OPTIONS")
+		api.HandleFunc("/experiments/{exp}/vms/{name}/unmount", UnmountVM).Methods("DELETE", "OPTIONS")
+		api.HandleFunc("/experiments/{exp}/vms/{name}/files", GetMountFiles).Methods("GET", "OPTIONS").Queries("path", "{path}")
+		api.HandleFunc("/experiments/{exp}/vms/{name}/files/download", DownloadMountFile).Methods("GET", "OPTIONS").Queries("path", "{path}")
+		api.HandleFunc("/experiments/{exp}/vms/{name}/files/upload", UploadMountFile).Methods("PUT", "OPTIONS").Queries("path", "{path}")
+	}
+
 	api.HandleFunc("/vms", GetAllVMs).Methods("GET", "OPTIONS")
 	api.HandleFunc("/applications", GetApplications).Methods("GET", "OPTIONS")
 	api.HandleFunc("/topologies", GetTopologies).Methods("GET", "OPTIONS")
--- a/src/go/api/config/default/experiment-viewer.yml
+++ b/src/go/api/config/default/experiment-viewer.yml
@@ -19,3 +19,8 @@ spec:
     - "*"
     verbs:
     - list
+  - resources:
+    - "vms/mount"
+    verbs:
+    - post
+    - delete
--- a/src/go/api/config/default/global-viewer.yml
+++ b/src/go/api/config/default/global-viewer.yml
@@ -10,6 +10,15 @@ spec:
     - "*/*"
     resourceNames:
     - "*"
+    - "*/*"
     verbs:
     - list
     - get
+  - resources:
+    - "vms/mount"
+    resourceNames:
+    - "*"
+    - "*/*"
+    verbs:
+    - post
+    - delete
--- a/src/go/api/config/default/vm-viewer.yml
+++ b/src/go/api/config/default/vm-viewer.yml
@@ -14,3 +14,10 @@ spec:
     - "vms/vnc"
     verbs:
     - get
+  - resources:
+    - "vms/mount"
+    verbs:
+    - post
+    - list
+    - delete
+    - get
--- a/src/go/api/experiment/experiment.go
+++ b/src/go/api/experiment/experiment.go
@@ -729,7 +729,7 @@ func Delete(name string) error {
 	return errors
 }
 
-func Files(name, filter string) (file.ExperimentFiles, error) {
+func Files(name, filter string) (file.Files, error) {
 	return file.GetExperimentFiles(name, filter)
 }
 
--- a/src/go/api/vm/vm.go
+++ b/src/go/api/vm/vm.go
@@ -99,6 +99,7 @@ func List(expName string) ([]mm.VM, error) {
 			vm.CPUs = details.CPUs
 			vm.RAM = details.RAM
 			vm.Disk = details.Disk
+			vm.CCActive = details.CCActive
 
 			// `vm.IPv4` could be nil/empty if minimega isn't reporting any IPs for it
 			if len(vm.IPv4) == 0 {
@@ -218,6 +219,7 @@ func Get(expName, vmName string) (*mm.VM, error) {
 	vm.CPUs = details[0].CPUs
 	vm.RAM = details[0].RAM
 	vm.Disk = details[0].Disk
+	vm.CCActive = details[0].CCActive
 
 	// `vm.IPv4` could be nil/empty if minimega isn't reporting any IPs for it
 	if len(vm.IPv4) == 0 {
--- a/src/go/app/startup.go
+++ b/src/go/app/startup.go
@@ -61,8 +61,9 @@ func (this Startup) PreStart(ctx context.Context, exp *types.Experiment) error {
 				if h, ok := ips[iface.Address()]; ok {
 					return fmt.Errorf("duplicate IP detected: %s and %s both have %s configured", h, node.General().Hostname(), iface.Address())
 				}
-
-				ips[iface.Address()] = node.General().Hostname()
+				if iface.Address() != "" {
+					ips[iface.Address()] = node.General().Hostname()
+				}
 			}
 		}
 
--- a/src/go/cmd/ui.go
+++ b/src/go/cmd/ui.go
@@ -57,6 +57,7 @@ func newUiCmd() *cobra.Command {
 				web.ServeWithTLS(viper.GetString("ui.tls-key"), viper.GetString("ui.tls-cert")),
 				web.ServePhenixLogs(viper.GetString("ui.logs.phenix-path")),
 				web.ServeMinimegaLogs(viper.GetString("ui.logs.minimega-path")),
+				web.ServeWithFeatures(viper.GetStringSlice("ui.features")),
 			}
 
 			if MustGetBool(cmd.Flags(), "log-requests") {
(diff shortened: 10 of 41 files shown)


activeshadow:
Dude @jacdavi this sounds awesome! Thank you! I'll test this and your other PR ASAP.

src/go/util/file/types.go
```diff
@@ -22,21 +24,22 @@ type ImageDetails struct {
 	Size     int
 }
 
-type ExperimentFile struct {
+type File struct {
 	Name       string   `json:"name"`
 	Path       string   `json:"path"`
 	Date       string   `json:"date"`
-	Size       int      `json:"size"`
+	Size       int64    `json:"size"`
```
activeshadow, line 31: Why the switch from `int` to `int64`? Is it just because `fs.FileInfo.Size` is an `int64`?

src/go/util/mm/minimega.go
```diff
@@ -148,6 +149,7 @@ func (this Minimega) GetVMInfo(opts ...Option) VMs {
 		vm.UUID = row["uuid"]
 		vm.Host = row["host"]
 		vm.Name = row["name"]
+		vm.CCActive, _ = strconv.ParseBool(row["cc_active"])
```
activeshadow, line 152: I have found the `cc_active` column from `vm info` to be unreliable. I've seen cases where a miniccc agent in a VM has failed for some reason and `cc_active` still reports true. I would suggest using the `mm.IsC2ClientActive` function instead, since it uses the more reliable `cc client` command to determine if cc is active for a VM.

src/go/web/handlers.go
```diff
@@ -3332,6 +3338,214 @@ func GetError(w http.ResponseWriter, r *http.Request) {
 	w.Write(body)
 }
 
+// POST /experiments/{exp}/vms/{name}/mount
+func MountVM(w http.ResponseWriter, r *http.Request) {
+	vars := mux.Vars(r)
+
+	role := r.Context().Value("role").(rbac.Role)
+	if !role.Allowed("vms/mount", "patch") {
```
activeshadow, line 3346: Should use `post` instead of `patch` here since we're creating a new mount, and should pass `{exp_name}/{vm_name}` to `role.Allowed` as well to support per-VM access control for mounting.

src/go/web/handlers.go
```diff
@@ -3332,6 +3338,214 @@ func GetError(w http.ResponseWriter, r *http.Request) {
...
+	
+	w.WriteHeader(http.StatusOK)
+}
+
+// POST /experiments/{exp}/vms/{name}/unmount
+func UnmountVM(w http.ResponseWriter, r *http.Request) {
+	vars := mux.Vars(r)
+
+	role := r.Context().Value("role").(rbac.Role)
+	if !role.Allowed("vms/mount", "patch") {
```
activeshadow, line 3372: Should use `delete` instead of `patch` here since we're deleting a mount, and should pass `{exp_name}/{vm_name}` to `role.Allowed` as well to support per-VM access control for unmounting.

src/go/web/handlers.go
```diff
@@ -3332,6 +3338,214 @@ func GetError(w http.ResponseWriter, r *http.Request) {
...
+	w.WriteHeader(http.StatusOK)
+}
+
+// GET /experiments/{exp}/vms/{name}/mount/files?path=
+func GetMountFiles(w http.ResponseWriter, r *http.Request) {
+	vars := mux.Vars(r)
+	basePath := mm.GetLocalMountPath(vars["exp"], vars["name"])
+
+	role := r.Context().Value("role").(rbac.Role)
+	if !role.Allowed("vms/mount", "patch") {
```
activeshadow, line 3398: Should use `list` instead of `patch` here since we're getting a list of mounted files, and should pass `{exp_name}/{vm_name}` to `role.Allowed` as well to support per-VM access control for mounts.

src/go/web/handlers.go
```diff
@@ -3332,6 +3338,214 @@ func GetError(w http.ResponseWriter, r *http.Request) {
...
+	w.Write(body)
+}
+
+// GET /experiments/{exp}/vms/{name}/files/download?path=
+func DownloadMountFile(w http.ResponseWriter, r *http.Request) {
+	vars := mux.Vars(r)
+	basePath := mm.GetLocalMountPath(vars["exp"], vars["name"])
+
+	role := r.Context().Value("role").(rbac.Role)
+	if !role.Allowed("vms/mount", "patch") {
```
activeshadow, line 3456: Should use `get` instead of `patch` here since we're downloading a specific file from a mount, and should pass `{exp_name}/{vm_name}` to `role.Allowed` as well to support per-VM access control for mounts.

src/go/web/handlers.go
```diff
@@ -3332,6 +3338,214 @@ func GetError(w http.ResponseWriter, r *http.Request) {
...
+	http.ServeFile(w, r, combinedPath)
+}
+
+// PUT /experiments/{exp}/vms/{name}/files/download?path=
+func UploadMountFile(w http.ResponseWriter, r *http.Request) {
+	vars := mux.Vars(r)
+	basePath := mm.GetLocalMountPath(vars["exp"], vars["name"])
+
+	role := r.Context().Value("role").(rbac.Role)
+	if !role.Allowed("vms/mount", "patch") {
```
activeshadow, line 3503: Should pass `{exp_name}/{vm_name}` to `role.Allowed` to support per-VM access control for mounts.

src/go/web/server.go
```diff
@@ -169,6 +169,13 @@ func Start(opts ...ServerOption) error {
 	api.HandleFunc("/experiments/{exp}/vms/{name}/snapshots/{snapshot}", RestoreVM).Methods("POST", "OPTIONS")
 	api.HandleFunc("/experiments/{exp}/vms/{name}/commit", CommitVM).Methods("POST", "OPTIONS")
 	api.HandleFunc("/experiments/{exp}/vms/{name}/memorySnapshot", CreateVMMemorySnapshot).Methods("POST", "OPTIONS")
+	api.HandleFunc("/experiments/{exp}/vms/{name}/mount", MountVM).Methods("POST", "OPTIONS")
+	api.HandleFunc("/experiments/{exp}/vms/{name}/unmount", UnmountVM).Methods("POST", "OPTIONS")
```
activeshadow, line 173: Let's change this from `POST` to `DELETE` for unmounting.

activeshadow (changes requested):
I still need to review the new/updated web handler code a bit more, as well as all the new/updated Vue code, so there may be additional requested changes in the future.

src/go/util/file/types.go
jacdavi: That's the main reason. Though it also seems like it doesn't hurt to be explicit on the int size since a 32 bit int would be problematic for files > 4.4GB.

src/go/web/handlers.go
jacdavi: Thanks for this. I wasn't really familiar with the RBAC system. 

One open question I need to figure out is that right now `GetMountFiles` only requires `list` permissions, but for that method to work, you'll have to have mounted the vm which requires `post`. 

My original thought was that only users should be able to use the feature, but maybe viewers should be able to create a mount and view files as they are. Any thoughts @aherna ?

jacdavi:
Also working on some fixes for multiple users/browsers interacting with mounts. Right now, a 2nd person cannot mount a vm if a user is currently downloading/uploading a file on any other vm. The best solution seems to be a per vm mutex.

Another issue is if two users are browsing the same vm and the 2nd closes the modal, that unmounts the vm breaking the 1st user's session when they try to do anything. Still thinking of the best way to prevent this. Maybe track user/sessions in the mount call and remove them in unmount. Then only actually run `cc mount` after there are no user/sessions remaining.

src/go/web/handlers.go
activeshadow: We can update the RBAC policies for the default roles we include in phenix to adjust for this. I agree that viewers should be able to mount and view files, and perhaps download, but not upload them.

As for the need to have both `post` and `list` permissions to actually list files, we can handle that in the policy config.

activeshadow:
> Also working on some fixes for multiple users/browsers interacting with mounts. Right now, a 2nd person cannot mount a vm if a user is currently downloading/uploading a file on any other vm. The best solution seems to be a per vm mutex.

Yes, I was going to suggest having a mutex per VM when I did a deeper review of the HTTP handlers code.
 
> Another issue is if two users are browsing the same vm and the 2nd closes the modal, that unmounts the vm breaking the 1st user's session when they try to do anything. Still thinking of the best way to prevent this. Maybe track user/sessions in the mount call and remove them in unmount. Then only actually run `cc mount` after there are no user/sessions remaining.

If it were me coding this, I would have probably gone down the route of having a resource counter per VM that got incremented/decremented as users opened the file browser modal. One of the issues I can think of here is to make sure the count gets decremented if a user browses away from the UI without explicitly closing the modal.

src/go/web/handlers.go
jacdavi: Alright, I think I've got this mostly done, though since I still don't fully know the RBAC system, I may have missed some stuff for the experiment/vm viewers.

activeshadow: Yeah, we can tease the RBAC stuff out as we go. The RBAC stuff on the UI side needs to be overhauled, since right now it's limited to the predefined roles. I know how I want to refactor it, just haven't had the time to do so yet.

activeshadow (changes requested):
Like you did in the vm-viewer.yml RBAC config file, I think you'll need to add the `list` and `get` verbs to the experiment-viewer.yml and global-viewer.yml RBAC config files for the `vms/mount` resource.

jacdavi:
> 

I did test setting myself to `global-viewer` on the backend and things seemed to work fine. Correct me if I'm wrong, but both already have star policies that should match for list/get.

activeshadow:
> I did test setting myself to `global-viewer` on the backend and things seemed to work fine. Correct me if I'm wrong, but both already have star policies that should match for list/get.

Whoops, you're right! I failed to review the other portions of those default policies before commenting.

jacdavi:
> > Also working on some fixes for multiple users/browsers interacting with mounts. Right now, a 2nd person cannot mount a vm if a user is currently downloading/uploading a file on any other vm. The best solution seems to be a per vm mutex.
> 
> Yes, I was going to suggest having a mutex per VM when I did a deeper review of the HTTP handlers code.
> 
> > Another issue is if two users are browsing the same vm and the 2nd closes the modal, that unmounts the vm breaking the 1st user's session when they try to do anything. Still thinking of the best way to prevent this. Maybe track user/sessions in the mount call and remove them in unmount. Then only actually run `cc mount` after there are no user/sessions remaining.
> 
> If it were me coding this, I would have probably gone down the route of having a resource counter per VM that got incremented/decremented as users opened the file browser modal. One of the issues I can think of here is to make sure the count gets decremented if a user browses away from the UI without explicitly closing the modal.

Ok, I made the change to a per-vm mutex and tracking the number of users who have mounted. Tracking unmounts is still somewhat unsolved. The user can only close the modal using the (x) and that correctly calls unmount. 

However, that doesn't account for closing the browser/tab or refreshing. I added a listener for `beforeunload`, but it seems inconsistent. It didn't work at all in Firefox and only worked on Chrome for refreshing.

src/go/util/mm/minimega.go
jacdavi: Let me know your thoughts on 280b16c135374abee536ec7963d3a08b64077b25. I found there was a fair bit of slow down if I ran `IsC2ClientActive` for every vm in an experiment, so I have it only running for single vm calls to `GetVMInfo` right now. It's possible I'm just not providing the right options and that's what is slowing things down.

The cc active flag is currently being used in the vm info modal and is displayed as text and used for disabling the mount button. I previously made the info modal make an extra backend call to get the latest info for its vm, so it should be getting the `IsC2ClientActive` value.

activeshadow: @jacdavi I'll take a look at this ASAP and see if I can't come up with a way to speed up client status requests for multiple VMs at the same time. I have an idea of how to do so.

activeshadow:
> > > Also working on some fixes for multiple users/browsers interacting with mounts. Right now, a 2nd person cannot mount a vm if a user is currently downloading/uploading a file on any other vm. The best solution seems to be a per vm mutex.
> > 
> > 
> > Yes, I was going to suggest having a mutex per VM when I did a deeper review of the HTTP handlers code.
> > > Another issue is if two users are browsing the same vm and the 2nd closes the modal, that unmounts the vm breaking the 1st user's session when they try to do anything. Still thinking of the best way to prevent this. Maybe track user/sessions in the mount call and remove them in unmount. Then only actually run `cc mount` after there are no user/sessions remaining.
> > 
> > 
> > If it were me coding this, I would have probably gone down the route of having a resource counter per VM that got incremented/decremented as users opened the file browser modal. One of the issues I can think of here is to make sure the count gets decremented if a user browses away from the UI without explicitly closing the modal.
> 
> Ok, I made the change to a per-vm mutex and tracking the number of users who have mounted. Tracking unmounts is still somewhat unsolved. The user can only close the modal using the (x) and that correctly calls unmount.
> 
> However, that doesn't account for closing the browser/tab or refreshing. I added a listener for `beforeunload`, but it seems inconsistent. It didn't work at all in Firefox and only worked on Chrome for refreshing.

@jacdavi ugh... okay I was afraid of that. It's great how all the browsers work so differently! /s

activeshadow:
@jacdavi have you tested this at all with phenix and minimega running in Docker? @glattercj and I are seeing some issues with files actually showing up in the file browser modal and we're assuming it has something to do with how the resulting mount from `cc mount` is being handled by Docker, but if you've tested it using Docker to run phenix and minimega then that wouldn't be our issue.

jacdavi:
> @jacdavi have you tested this at all with phenix and minimega running in Docker? @glattercj and I are seeing some issues with files actually showing up in the file browser modal and we're assuming it has something to do with how the resulting mount from `cc mount` is being handled by Docker, but if you've tested it using Docker to run phenix and minimega then that wouldn't be our issue.

I haven't done any testing with docker. I have seen a few issues with individual files (e.g., can't download, wrong size), but those issues also existed when using minimega from the CLI. So due to time constraints, I considered them out of scope. Is this something similar or do the problems only occur with the new GUI?

activeshadow:
> > @jacdavi have you tested this at all with phenix and minimega running in Docker? @glattercj and I are seeing some issues with files actually showing up in the file browser modal and we're assuming it has something to do with how the resulting mount from `cc mount` is being handled by Docker, but if you've tested it using Docker to run phenix and minimega then that wouldn't be our issue.
> 
> I haven't done any testing with docker. I have seen a few issues with individual files (e.g., can't download, wrong size), but those issues also existed when using minimega from the CLI. So due to time constraints, I considered them out of scope. Is this something similar or do the problems only occur with the new GUI?

@jacdavi I just tested again using Docker containers to run minimega and phenix and it does seem to be working as expected now. I'm going to do some additional review and testing today to see if we can get a bow on this and get it merged in. This is awesome!

activeshadow (changes requested):
image

Can we widen the modal a little so the buttons at the bottom are all on the same row?

I'm also wondering if we need `CC Active` status in this modal. I've been thinking of how we could highlight whether or not CC is active for a VM in the main VM table so users don't have to click into the VM modal to know the status.

activeshadow:
@jacdavi I noticed an edge case where I killed the miniccc agent in a VM, but the UI still saw it as active so it let me attempt to mount the VM's disk. After attempting to mount, phenix became unresponsive for any actions that required a call to minimega, and eventually I got a toast saying something about it not being able to unmount the VM.

src/go/util/mm/minimega.go
```diff
@@ -761,6 +762,20 @@ func (Minimega) IsC2ClientActive(opts ...C2Option) error {
...
+	o := NewC2Options(opts...)
+	if o.skipActiveClientCheck {
+		return nil
+	}
+
+	vms := GetVMInfo(NS(o.ns), VMName(o.vm))
+	if len(vms) == 0 {
+		return fmt.Errorf("VM %s does not exist", o.vm)
+	}
+	return mm.isC2ClientActive(vms[0], opts...)
```
activeshadow, line 776: @jacdavi why did you break this into two separate functions?

jacdavi: Yeah exactly, not the most elegant though. For putting `cc active` in the table, we'd definitely want the better check in place for a list call to `GetVmInfo` as opposed to just limiting to one vm.

activeshadow:
I think we need to update the `RunningExperiment` UI component to do an extra call to get the VM info when the VM details modal is opened so we can get an up-to-date status of the CC active state. Right now, the CC active state is grabbed only when the running experiment table is first loaded. So if someone is on that table for a while it could get stale.

In general I think it would be nice to also periodically update the entire table to keep the uptime column updated and the CC active status updated if we show that in the table view. Perhaps this is ripe for a future effort though.

jacdavi:
> @jacdavi I noticed an edge case where I killed the miniccc agent in a VM, but the UI still saw it as active so it let me attempt to mount the VM's disk. After attempting to mount, phenix became unresponsive for any actions that required a call to minimega, and eventually I got a toast saying something about it not being able to unmount the VM.

So this is probably the call to `mm.ExecC2Command` hanging since I didn't provide a timeout and I think the default is 5m. What do you think a sensible timeout would be? Also, the call to `mm.ExecC2Command` should call `IsC2ClientActive` before continuing, so then that method might not have caught that cc was inactive.

jacdavi:
> I think we need to update the `RunningExperiment` UI component to do an extra call to get the VM info when the VM details modal is opened so we can get an up-to-date status of the CC active state. Right now, the CC active state is grabbed only when the running experiment table is first loaded. So if someone is on that table for a while it could get stale.
> 
> In general I think it would be nice to also periodically update the entire table to keep the uptime column updated and the CC active status updated if we show that in the table view. Perhaps this is ripe for a future effort though.

I did add a call to update the modal when it's opened: https://github.com/sandia-minimega/phenix/blob/180a1c5bda418ed5fbb473baed04800142dc406d/src/js/src/components/RunningExperiment.vue#L1590
Otherwise, when you started an experiment, it would never show cc as active until you reloaded the page.

Agree about the periodic update!

activeshadow:
> > I think we need to update the `RunningExperiment` UI component to do an extra call to get the VM info when the VM details modal is opened so we can get an up-to-date status of the CC active state. Right now, the CC active state is grabbed only when the running experiment table is first loaded. So if someone is on that table for a while it could get stale.
> > In general I think it would be nice to also periodically update the entire table to keep the uptime column updated and the CC active status updated if we show that in the table view. Perhaps this is ripe for a future effort though.
> 
> I did add a call to update the modal when it's opened:
> 
> https://github.com/sandia-minimega/phenix/blob/180a1c5bda418ed5fbb473baed04800142dc406d/src/js/src/components/RunningExperiment.vue#L1590
> 
> 
> Otherwise, when you started an experiment, it would never show cc as active until you reloaded the page.
> Agree about the periodic update!

@jacdavi oh sweet! I wonder why I still saw that edge case then where I killed the miniccc agent in a VM but it still showed as active when I opened the modal. I'll test it again to see if I can reproduce. Thanks!

activeshadow:
> > @jacdavi I noticed an edge case where I killed the miniccc agent in a VM, but the UI still saw it as active so it let me attempt to mount the VM's disk. After attempting to mount, phenix became unresponsive for any actions that required a call to minimega, and eventually I got a toast saying something about it not being able to unmount the VM.
> 
> So this is probably the call to `mm.ExecC2Command` hanging since I didn't provide a timeout and I think the default is 5m. What do you think a sensible timeout would be? Also, the call to `mm.ExecC2Command` should call `IsC2ClientActive` before continuing, so then that method might not have caught that cc was inactive.

This is a good lead to go down... I'll take a look now. Thanks!

src/go/util/mm/minimega.go
activeshadow: I think I came up with a way to speed this up. It seems to be working as expected.

activeshadow:
> > > @jacdavi I noticed an edge case where I killed the miniccc agent in a VM, but the UI still saw it as active so it let me attempt to mount the VM's disk. After attempting to mount, phenix became unresponsive for any actions that required a call to minimega, and eventually I got a toast saying something about it not being able to unmount the VM.
> > 
> > 
> > So this is probably the call to `mm.ExecC2Command` hanging since I didn't provide a timeout and I think the default is 5m. What do you think a sensible timeout would be? Also, the call to `mm.ExecC2Command` should call `IsC2ClientActive` before continuing, so then that method might not have caught that cc was inactive.
> 
> This is a good lead to go down... I'll take a look now. Thanks!

@jacdavi okay so I updated the mount/unmount calls to use a timeout and that seems to help with the edge case. The only issue I'm hitting now is Docker specific... if I try to mount a VM that doesn't have cc active (but the UI still thinks it does), it times out as I mentioned but when I try to shut down the minimega container Docker fails to kill it. Perhaps we can address this issue at a different time since it's specific to Docker.

activeshadow:
@jacdavi at some point soon I'd like to rebase your branch on the current master, but that will mean doing a force push to your branch. Let me know when you're okay with me doing that since you'll have to delete your branch locally and then pull it down again from GitHub.

jacdavi:
> @jacdavi at some point soon I'd like to rebase your branch on the current master, but that will mean doing a force push to your branch. Let me know when you're okay with me doing that since you'll have to delete your branch locally and then pull it down again from GitHub.

I'm good whenever you are :). Both of your commits look great

activeshadow:
> > > > Also working on some fixes for multiple users/browsers interacting with mounts. Right now, a 2nd person cannot mount a vm if a user is currently downloading/uploading a file on any other vm. The best solution seems to be a per vm mutex.
> > > 
> > > 
> > > Yes, I was going to suggest having a mutex per VM when I did a deeper review of the HTTP handlers code.
> > > > Another issue is if two users are browsing the same vm and the 2nd closes the modal, that unmounts the vm breaking the 1st user's session when they try to do anything. Still thinking of the best way to prevent this. Maybe track user/sessions in the mount call and remove them in unmount. Then only actually run `cc mount` after there are no user/sessions remaining.
> > > 
> > > 
> > > If it were me coding this, I would have probably gone down the route of having a resource counter per VM that got incremented/decremented as users opened the file browser modal. One of the issues I can think of here is to make sure the count gets decremented if a user browses away from the UI without explicitly closing the modal.
> > 
> > 
> > Ok, I made the change to a per-vm mutex and tracking the number of users who have mounted. Tracking unmounts is still somewhat unsolved. The user can only close the modal using the (x) and that correctly calls unmount.
> > However, that doesn't account for closing the browser/tab or refreshing. I added a listener for `beforeunload`, but it seems inconsistent. It didn't work at all in Firefox and only worked on Chrome for refreshing.
> 
> @jacdavi ugh... okay I was afraid of that. It's great how all the browsers work so differently! /s

@jacdavi given how you wrote the mount/unmount code, I don't think it's too big of a deal if a user browses away from the UI with the file browser modal open. Yes, the mount will end up staying active until the experiment is stopped and the VM is shut down but I think that's acceptable.

jacdavi:
> > > > > Also working on some fixes for multiple users/browsers interacting with mounts. Right now, a 2nd person cannot mount a vm if a user is currently downloading/uploading a file on any other vm. The best solution seems to be a per vm mutex.
> > > > 
> > > > 
> > > > Yes, I was going to suggest having a mutex per VM when I did a deeper review of the HTTP handlers code.
> > > > > Another issue is if two users are browsing the same vm and the 2nd closes the modal, that unmounts the vm breaking the 1st user's session when they try to do anything. Still thinking of the best way to prevent this. Maybe track user/sessions in the mount call and remove them in unmount. Then only actually run `cc mount` after there are no user/sessions remaining.
> > > > 
> > > > 
> > > > If it were me coding this, I would have probably gone down the route of having a resource counter per VM that got incremented/decremented as users opened the file browser modal. One of the issues I can think of here is to make sure the count gets decremented if a user browses away from the UI without explicitly closing the modal.
> > > 
> > > 
> > > Ok, I made the change to a per-vm mutex and tracking the number of users who have mounted. Tracking unmounts is still somewhat unsolved. The user can only close the modal using the (x) and that correctly calls unmount.
> > > However, that doesn't account for closing the browser/tab or refreshing. I added a listener for `beforeunload`, but it seems inconsistent. It didn't work at all in Firefox and only worked on Chrome for refreshing.
> > 
> > 
> > @jacdavi ugh... okay I was afraid of that. It's great how all the browsers work so differently! /s
> 
> @jacdavi given how you wrote the mount/unmount code, I don't think it's too big of a deal if a user browses away from the UI with the file browser modal open. Yes, the mount will end up staying active until the experiment is stopped and the VM is shut down but I think that's acceptable.

Ultimately I agree. Plus minimega will clear any mounts when you kill a VM. So shutting down the VM/experiment in phenix will clean everything up. 

I did discover that things hang if you "pause" a vm and try to mount it. However, the same is true if you "stop" a vm in minimega and try to `ls` the mount directory. So I'm planning to open a bug for minimega on that one.

activeshadow:
Okay @jacdavi I just did a force push. Please get your local branch in sync and do another round of testing to make sure the merge conflicts I fixed didn't end up breaking anything. All of the conflicts were in the UI javascript code.

activeshadow:
I can see issues arising when, for example, a Windows VM that has to reboot a couple of times gets mounted by a user in between reboots. I just had this happen to me while testing. This is partly caused by using the VM UUID to detect if the cc agent is active, but using the host name won't always prevent the issue. This is also partly caused by minimega's default frequency for reaping dead cc agents being 30s, but again, having it happen more often won't help prevent all edge cases.

activeshadow:
Also, I think I may have borked the Scorch table when addressing merge conflicts.

jacdavi:
> Also, I think I may have borked the Scorch table when addressing merge conflicts.

I might not have updated that buefy table correctly. I've never used scorch, so I wasn't sure how to check it...

activeshadow:
@jacdavi I think this might be ready... however, I'd like to look into putting the feature behind a feature flag just in case it's causing issues for someone. If you want to go ahead and do one more round of testing with the latest commits that would be great. I can then squash all the commits and see where I'm at with the feature flag stuff.

jacdavi:
> @jacdavi I think this might be ready... however, I'd like to look into putting the feature behind a feature flag just in case it's causing issues for someone. If you want to go ahead and do one more round of testing with the latest commits that would be great. I can then squash all the commits and see where I'm at with the feature flag stuff.

I like the feature flag, that makes a lot of sense. Is there some type of documentation where it would be appropriate to mention the flag? Also thanks, for separating the mount functions to their own file. I wanted to do that, but wasn't sure if there was a convention I would be breaking.

Tested things out again, and they look good to me!

aherna:
> I can see issues arising when, for example, a Windows VM that has to reboot a couple of times gets mounted by a user in between reboots. I just had this happen to me while testing. This is partly caused by using the VM UUID to detect if the cc agent is active, but using the host name won't always prevent the issue. This is also partly caused by minimega's default frequency for reaping dead cc agents being 30s, but again, having it happen more often won't help prevent all edge cases.

We need to make sure uuid is still used as hostname with our experiments is not workable. Hostnames are dynamically generated at boot time and wont know what they will be.

activeshadow:
> > I can see issues arising when, for example, a Windows VM that has to reboot a couple of times gets mounted by a user in between reboots. I just had this happen to me while testing. This is partly caused by using the VM UUID to detect if the cc agent is active, but using the host name won't always prevent the issue. This is also partly caused by minimega's default frequency for reaping dead cc agents being 30s, but again, having it happen more often won't help prevent all edge cases.
> 
> @jacdavi @activeshadow We need to make sure uuid is still used as hostname with our experiments is not workable. Hostnames are dynamically generated at boot time and wont know what they will be.
> 
> we use uuid for soh

@aherna the convention we use for phenix in almost all cases but yours is to use the host name to tell when cc is ready because Windows VMs have typically gone through all their reboots once the host name reported by miniccc is the same as the host name used in the topology file.

SoH, by default, uses host names. You must set the option to use UUIDs instead.

I'm wondering if it would make sense to add a setting to the topology node schema that signifies whether or not the host name of the VM will end up matching the host name used in the topology. If not, then we'd know to use UUIDs for that particular VM.

I'm open to other ideas as well, but I will keep pushing against using UUIDs (by default anyway) when determining if cc is active for VMs, especially when it comes to allowing a VM's file system to be mounted.

jacdavi:
> > > I can see issues arising when, for example, a Windows VM that has to reboot a couple of times gets mounted by a user in between reboots. I just had this happen to me while testing. This is partly caused by using the VM UUID to detect if the cc agent is active, but using the host name won't always prevent the issue. This is also partly caused by minimega's default frequency for reaping dead cc agents being 30s, but again, having it happen more often won't help prevent all edge cases.
> > 
> > 
> > @jacdavi @activeshadow We need to make sure uuid is still used as hostname with our experiments is not workable. Hostnames are dynamically generated at boot time and wont know what they will be.
> > we use uuid for soh
> 
> @aherna the convention we use for phenix in almost all cases but yours is to use the host name to tell when cc is ready because Windows VMs have typically gone through all their reboots once the host name reported by miniccc is the same as the host name used in the topology file.
> 
> SoH, by default, uses host names. You must set the option to use UUIDs instead.
> 
> I'm wondering if it would make sense to add a setting to the topology node schema that signifies whether or not the host name of the VM will end up matching the host name used in the topology. If not, then we'd know to use UUIDs for that particular VM.
> 
> I'm open to other ideas as well, but I will keep pushing against using UUIDs (by default anyway) when determining if cc is active for VMs, especially when it comes to allowing a VM's file system to be mounted.

Yeah Arthur is right that mounting won't work with our Windows VMs. I see you changed the cc check, but we'd also need to change the actual mount/unmount exec commands.

If you'd rather not use UUID for the general release, would it make sense for me to make another fork with UUIDs for our setup use? Would that reintroduce any bugs you fixed? Then the PR can be merged as-is or with the schema change you suggested.

jacdavi:
I added 6063b0b547ae4f00b2c5ee45e4dae69d25544e96 since I was unable to start up our experiment that has a number of interfaces without an address like below. I can undo it if this is an issue with our setup.

```
          network:
            interfaces:
                - name: eth0
                  proto: dhcp
                  type: ethernet
                  vlan: vlan
```

activeshadow:
> I added [6063b0b](https://github.com/sandia-minimega/phenix/commit/6063b0b547ae4f00b2c5ee45e4dae69d25544e96) since I was unable to start up our experiment that has a number of interfaces without an address like below. I can undo it if this is an issue with our setup.
> 
> ```
>           network:
>             interfaces:
>                 - name: eth0
>                   proto: dhcp
>                   type: ethernet
>                   vlan: vlan
> ```

Ah yes, good catch! We don't use DHCP much, as you have probably noticed. :joy:

activeshadow (approved):

activeshadow:
@jacdavi I did some additional testing with auth enabled and on a URL path that isn't `/` and everything seemed to work as expected. As such, I think I'm ready to squash and merge this in. Let me know how testing has gone for you today.

jacdavi:
Things look good to me, I'm ready for the merge!

PR #115, opened 2022-12-17
update noVNC to 1.3.0

jacdavi:
Updates noVNC to 1.3.0 to match recent [minimega update](https://github.com/sandia-minimega/minimega/pull/1488)

I think I updated everything phenix specific (e.g., page banner and title). Did noVNC ever work when running the dev webserver (`npm run serve`)?


activeshadow:
@jacdavi did you happen to test the VM banner capability with this branch yet?

jacdavi:
> @jacdavi did you happen to test the VM banner capability with this branch yet?

Yeah, I did confirm that the banner still shows. It's not readable for the first second since noVNC displays a toast about connecting to the VM, but that's probably not a big enough deal to worry about.

activeshadow:
@jacdavi the only issue I see with banners is that the top one is allowed to be scrolled away in some instances.

activeshadow:
> @jacdavi the only issue I see with banners is that the top one is allowed to be scrolled away in some instances.

@jacdavi I found a fix so I'll add it in and then merge this. Thank you!

jacdavi:
Thanks @activeshadow !

PR #118, opened 2023-01-21
Add support for dns in topology config

glattercj:
E.g.,

```yaml
...
    network:
      interfaces:
      - name: foo
        address: 192.168.1.1
        gateway: 192.168.254
        dns: 192.168.1.99
...
```


activeshadow:
Good stuff @glattercj thank you!

What do you think about making the DNS entry an array so multiple servers can be specified?

We also need to add it to the UI Builder schema. Too many schemas to have to update... 🤔😂 (that's my fault).

activeshadow:
@glattercj why did you close this?

Don't merge it yet because I have some additional changes to suggest but please do reopen it!

glattercj:
Making some updates.

PR #119, opened 2023-01-22
Add support for DNS in topology

glattercj:
Supports either string or array format, e.g.:

```yaml
...
    network:
      interfaces:
      - name: foo
        address: 192.168.1.1
        gateway: 192.168.254
        dns: 192.168.1.99
...
```

or

```yaml
...
    network:
      interfaces:
      - name: foo
        address: 192.168.1.1
        gateway: 192.168.254
        dns:
        - 192.168.1.99
        - 192.168.2.99
...
```


src/go/util/string.go
```diff
@@ -19,3 +21,32 @@ func RandomString(n int) string {
...
+// Allow yaml parsing for an array of strings or a single inlined element:
+//
+//	field: value
+//
+//	  or
+//
+//	field:
+//	  - value1
+//	  - value2
+type StringArray []string
```
activeshadow, line 36: Um... no. :joy:

Let's **strongly** consider naming this `StringSlice` instead please!

src/go/util/string.go
```diff
@@ -19,3 +21,32 @@ func RandomString(n int) string {
...
+		err := value.Decode(&single)
+		if err != nil {
+			return err
+		}
+		*a = []string{single}
+	} else {
+		*a = multi
+	}
+	return nil
+}
```
activeshadow, line 52: Will likely need to have a custom implementation of `UnmarshalJSON` as well.

src/go/tmpl/templates/windows_startup.tmpl
```diff
@@ -88,6 +88,13 @@ $wmi[{{ $idx }}].SetGateways('{{ $iface.Gateway }}', 1) | Out-Null
 $wmi.SetGateways('{{ $iface.Gateway }}', 1) | Out-Null
         {{ end }}
     {{ end }}
+    {{ if $iface.DNS }}
+        {{ if gt $length 1 }}
+$wmi[{{ $idx }}].SetDNSServerSearchOrder(@('{{ stringsJoin $iface.DNS "', '" }}')) | Out-Null
+        {{ else }}
+$wmi.SetDNSServerSearchOrder(@('{{ stringsJoin $iface.DNS "', '" }}')) | Out-Null
+        {{ end }}
+    {{ end }}
```
activeshadow, line 97: We also set the DNS server address below if this Windows VM is being added to a domain controller. How should we handle the case where DNS servers are provided for an interface, and the VM is being added to a domain controller, and the DNS servers provided are not the domain controller?

In order for AD to work across subnets, the interface that can reach the domain controller must use the domain controller as its DNS server. So, I'm thinking this might work as written, and if this VM is being added to a domain controller then its DNS server entry will just get overwritten below. We should probably document this procedure though, so no one gets confused if the DNS server they enter doesn't match.

activeshadow (changes requested):

src/go/util/string.go
glattercj: How about `type A []string`

src/go/util/string.go
glattercj: Without any additional changes, I tested with:

```bash
phenix cfg get topology/foobar -o json -p > foobar.json
phenix cfg delete topology/foobar
phenix cfg create /phenix/topologies/foobar.json
```

Output was: `The Topology/foobar configuration was created`, and the yaml shown with `cfg get` was correct.

Where else should I check?

glattercj: `topology/foobar` had 2 vms: one with dns as a string, one as an array with 2 entries.

activeshadow: @glattercj getting the JSON from phenix will output an array of DNS names. Did you test creating from JSON when the DNS entry was a string instead of an array of strings?

activeshadow: > `topology/foobar` had 2 vms: one with dns as a string, one as an array with 2 entries.

Was the DNS entry for the one VM a string when you ran `cfg create` using the JSON file (your last step)? When you export from phenix to JSON, it will be an array no matter what since it's a string slice in the topology schema.

glattercj: [foo-yaml.md](https://github.com/sandia-minimega/phenix/files/10492626/foo-yaml.md)
[foo-json.md](https://github.com/sandia-minimega/phenix/files/10492625/foo-json.md)

glattercj: Getting from phenix the json matched: one vm was a string, one was an array with 2 entries. It seems to be an array when it gets added as an experiment, then the exp config shows both as arrays.

activeshadow: Cool! Looks like the JSON parser handles that situation automatically. 👍🏻👍🏻

activeshadow: @glattercj I'm assuming you added the custom yaml unmarshaller because it wouldn't work by default?

activeshadow: @glattercj I just did some testing and you don't need the custom `StringSlice` type with the custom `UnmarshalYAML` function. I think everything works fine now that you updated the schema to allow for string or array.

glattercj: Fixed

src/go/tmpl/templates/windows_startup.tmpl
activeshadow: @glattercj any thoughts on this, or should we just leave it as-is?

glattercj: I think it's fine. Was going to test but don't have a DC topo handy. The main use case for this was external access for a VM, so folks (hopefully) know what they're doing if they also have configured a DC in the scenario config.

activeshadow (approved):

src/go/types/version/v1/network.go
```diff
@@ -254,6 +259,12 @@ func (this *Interface) SetGateway(gw string) {
 	this.GatewayF = gw
 }
 
+func (this *Interface) SetDNS(dns []string) {
+	for _, value := range dns {
+		this.DNSF = append(this.DNSF, value)
+	}
```
activeshadow, line 265: @glattercj now that we're using `[]string` instead of `StringSlice`, you can just do `this.DNSF = dns` here. Will save us a for-loop, and keep it consistent with other setters.

activeshadow (changes requested):
On more change... :joy:

src/go/types/version/v1/network.go
glattercj: Rgr

PR #127, opened 2023-05-29
CommitToDisk:Remove hardcoded minimega base directory

eric-c-wood:
The minimega base directory was hardcoded in the `CommitToDisk` function which would cause the UI `Create Backing Image` to fail when the user specifies the `--base-dir.minimega` option on the command line.


activeshadow:
Thank you @eric-c-wood!

PR #129, opened 2023-06-13
Record basic information as artifact for each SCORCH run

activeshadow:
Each scorch run creates a `/phenix/images/{{BRANCH_NAME}}/files/scorch/run-#/info-scorch-run#_time.txt` file which records basic information about the scorch run, including: Experiment Name, Scorch Run Name, Scorch Run Start Time, Scorch Run Stop Time, Phenix Version, Minimega Version.

### DIFF
--- a/src/go/api/scorch/app.go
+++ b/src/go/api/scorch/app.go
@@ -18,7 +18,9 @@ import (
 	"phenix/types"
 	ifaces "phenix/types/interfaces"
 	"phenix/util"
+	"phenix/util/mm/mmcli"
 	"phenix/util/shell"
+	"phenix/version"
 	"phenix/web/scorch"
 
 	log "github.com/activeshadow/libminimega/minilog"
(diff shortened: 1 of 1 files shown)


activeshadow:
@nblair2 looks good! Do you mind moving the code you added to a separate function named something like `recordInfo` so we can keep it as a separate unit of code?

I'll have some additional suggestions once it's moved.

Thanks for this!

nblair2:
@activeshadow refactored - please take a look and let me know if I am using too many args for the function.

src/go/api/scorch/app.go
```diff
@@ -326,6 +332,38 @@ func (this Scorch) stopFilebeat(ctx context.Context, cmd *exec.Cmd, port int) {
 	}
 }
 
+func (this Scorch) recordInfo(runID int, runDir string, name string, startTime time.Time, versionCommit string, versionTag string, versionDate string) error {
```
activeshadow, line 335: @nblair2 you don't need to pass all the version args here, since you can just use the version package to grab them directly in this function. Also, when two function args are of the same type, you can shorten the arg signature a little bit. In this case, you can do the following:

`recordInfo(runID int, runDir, name string, startTime time.Time) error`

src/go/api/scorch/app.go
```diff
@@ -326,6 +332,38 @@ func (this Scorch) stopFilebeat(ctx context.Context, cmd *exec.Cmd, port int) {
...
+	// runID, runDir, exp.Metadata.Name, start, version.Commit, version.Tag version.Date
+
+	var (
+		c    *mmcli.Command
+		resp string
+		err  error
+		info string
+	)
```
activeshadow, line 343: @nblair2 there's no need to define these variables ahead of time.

src/go/api/scorch/app.go
```diff
@@ -326,6 +332,38 @@ func (this Scorch) stopFilebeat(ctx context.Context, cmd *exec.Cmd, port int) {
...
+	c = mmcli.NewCommand()
+	c.Command = "version"
+	if resp, err = mmcli.SingleResponse(mmcli.Run(c)); err != nil {
+		return fmt.Errorf("running minimega verssion %w", err)
+	}
+
+	info = fmt.Sprintf("Experiment Name: %s\nExperiment Commit Hash: pending\nScorch Run Name: %s\nStart Time: %v\nEnd Time: %v\nPhenix Version: %s %s %s \nMinimega Version: %s",
+		name, this.md.RunName(runID), startTime, time.Now().UTC(), versionCommit, versionTag, versionDate, resp)
+
+	if _, err = os.Stat(runDir); os.IsNotExist(err) {
```
activeshadow, line 354: @nblair2 you don't really need to check that the directory exists before making it with `os.MkdirAll`, since `os.MkdirAll` will be a no-op if it already exists (no error will be returned).

src/go/api/scorch/app.go
```diff
@@ -326,6 +332,38 @@ func (this Scorch) stopFilebeat(ctx context.Context, cmd *exec.Cmd, port int) {
...
+		if err = os.MkdirAll(runDir, 0755); err != nil {
+			return fmt.Errorf("creating %s directory for scorch run %d: %w", runDir, runID, err)
+		}
+	}
+
+	if err := os.WriteFile(filepath.Join(runDir, fmt.Sprintf("info-scorch-run-%d_%s.txt", runID, startTime.Format(time.RFC3339))), []byte(info), 0666); err != nil {
```
activeshadow, line 360: @nblair2 since you use `fmt.Sprintf("info-scorch-run-%d_%s.txt", runID, startTime)` here and in the error message below, you might consider just pulling it out into its own variable so that 1) the formatted string only has to be generated once, and 2) readability is a little bit better.

activeshadow (commented):
@nblair2 just a couple of nit-picky things from my end...

nblair2:
@activeshadow - I appreciate the pointers. Changes made.

activeshadow:
Closing this so @nblair2 can open a new PR for this himself.

PR #132, opened 2023-07-03
feat: allow users to specify external nodes in topology

activeshadow:
It may be the case that a user would want to include nodes in the topology that don't necessarily need to be deployed by phenix in minimega. For example, lets say a user app exists to deploy some topology nodes in a cloud provider. Or, lets say the state of health app needs to know about hardware-in-the-loop nodes to do custom connectivity tests against.

phenix now supports `external` nodes that are not deployed in minimega, but still show up in the list of nodes in a topology, still show up in the state of health view, etc.

A node can be marked as external by adding the top-level `external: true` setting to a node's configuration in the topology configuration file.


src/go/types/version/v2/schema.go
```diff
@@ -195,8 +196,7 @@ components:
           $ref: "#/components/schemas/Topology"
         scenario:
           nullable: true
```
beleyj, line 198: We are unable to launch any experiment that does not have a scenario file. Based on our knowledge, scheme v1 is used for launching experiments and not v2. We see in v2 you have `nullable: true` for scenarios. However, you do not have `nullable: true` in schema v1 for `scenario`. It is very possible this null bug is from an old branch and not this one specific.

*Error:*
`
creating experiment config: validating config after config hooks: config validation failed: Error at "/scenario": Value is not nullable Schema: { "properties": { "apps": { "items": { "properties": { "assetDir": { "example": "/phenix/topologies/example-topo/assets", "type": "string" }, "hosts": { "items": { "properties": { "hostname": { "example": "example-host", "type": "string" }, "metadata": { "additionalProperties": true, "example": { "setting0": true, "setting1": 42, "setting2": "universe key" }, "nullable": true, "type": "object" } }, "required": [ "hostname" ], "type": "object" }, "type": "array" }, "metadata": { "additionalProperties": true, "example": { "setting0": true, "setting1": 42, "setting2": "universe key" }, "nullable": true, "type": "object" }, "name": { "example": "example-app", "type": "string" } }, "required": [ "name" ], "type": "object" }, "type": "array" } }, "required": [ "apps" ], "type": "object" } Value: null
`

PR #133, opened 2023-07-03
feat(scorch): add state of health component

activeshadow:
List of checks that state of health can be limited to via the `soh` scorch component (using the `checks` metadata setting):

* network-config
* reachability
* custom-reachability
* processes
* ports
* custom
* cpu-load
* flows

The other metadata settings are `failOnError` and `logLevel`.


activeshadow:
@glattercj and @mgaliar this is a pretty hefty PR so I would appreciate a review from one or both of you, even if it's a quick one.

In addition, I have some documentation staged in the dev version of the official docs for this. You can see them [here](https://phenix.sceptre.dev/dev/scorch/#soh-component).

mgaliar:
@activeshadow I have only had a chance to minimally look at this so far. First thing I tested was the SOAP soh to make sure that still worked as expected. I actually got an error with the mirror app. I think it probably has to do with the update from go 1.18 to 1.20. 
`panic: Something in this program imports go4.org/unsafe/assume-no-moving-gc to declare that it assumes a non-moving garbage collector, but your version of go4.org/unsafe/assume-no-moving-gc hasn't been updated to assert that it's safe against the go1.20 runtime. If you want to risk it, run with environment variable ASSUME_NO_MOVING_GC_UNSAFE_RISK_IT_WITH=go1.20 set. Notably, if go1.20 adds a moving garbage collector, this program is unsafe to use.goroutine 1 [running]:go4.org/unsafe/assume-no-moving-gc.init.0()      go4.org/unsafe/assume-no-moving-gc@v0.0.0-20220617031537-928513b29760/untested.go:25 +0x1ba`

activeshadow:
Thanks @mgaliar for taking a look. This is an easy fix, and I will try to push it ASAP so you can test again.

More to come.

activeshadow:
@mgaliar okay try this again when you get a chance. Thanks!

mgaliar:
@activeshadow That fixes it. And I looked at the rest of the PR and looks good to me!

mgaliar (approved):

glattercj (approved):
LGTM

PR #136, opened 2023-07-17
feat(workflow): allow API to accept tags for experiments

nblair2:
Users can now send key=value pairs via URL requests to the API when
applying a configuration file. Currently, any query in the form
'tag=key=value' will be stored as an experiment annotation named
`phenix.workflow/tags`, with the value being a simple coma-separated
string of the tags.

Take the following URL as an example.

```
/api/v1/workflow/apply/{BRANCH}?tag=key1=value1&tag=key2=value&foo=bar
```

This will result in the following annotation being added to the
experiment metadata.

```
metadata:
  annotations:
    "phenix.workflow/tags": "key1=value1,key2=value2"
```

Note that the `foo=bar` query parameter is ignored. These tags are then
recorded with other info after each scorch run.

### DIFF
--- a/src/go/api/scorch/app.go
+++ b/src/go/api/scorch/app.go
@@ -16,6 +16,7 @@ import (
 	"phenix/api/scorch/scorchexe"
 	"phenix/api/scorch/scorchmd"
 	"phenix/app"
+	"phenix/store"
 	"phenix/types"
 	ifaces "phenix/types/interfaces"
 	"phenix/util"
@@ -159,7 +160,7 @@ func (this *Scorch) Running(ctx context.Context, exp *types.Experiment) error {
 		this.stopFilebeat(ctx, cmd, port)
 	}
 
-	if err := this.recordInfo(runID, runDir, exp.Metadata.Name, start); err != nil {
+	if err := this.recordInfo(runID, runDir, exp.Metadata, start); err != nil {
 		errors = multierror.Append(errors, err)
 	}
 
@@ -333,7 +334,7 @@ func (this Scorch) stopFilebeat(ctx context.Context, cmd *exec.Cmd, port int) {
 	}
 }
 
-func (this Scorch) recordInfo(runID int, runDir, name string, startTime time.Time) error {
+func (this Scorch) recordInfo(runID int, runDir string, md store.ConfigMetadata, startTime time.Time) error {
 	c := mmcli.NewCommand()
 	c.Command = "version"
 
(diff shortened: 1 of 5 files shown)


src/go/api/scorch/app.go
```diff
@@ -159,7 +159,7 @@ func (this *Scorch) Running(ctx context.Context, exp *types.Experiment) error {
 		this.stopFilebeat(ctx, cmd, port)
 	}
 
-	if err := this.recordInfo(runID, runDir, exp.Metadata.Name, start); err != nil {
+	if err := this.recordInfo(runID, runDir, exp.Metadata.Name, exp.Metadata.Annotations["phenix.workflow/tags"], start); err != nil {
```
activeshadow, line 162: Since we're now passing more than one experiment metadata related variable, it might make sense to just pass a single experiment metadata variable to the `recordInfo` function and let it pull the relevant keys out of the metadata.

src/go/web/workflow.go
```diff
@@ -32,8 +32,17 @@ func ApplyWorkflow(w http.ResponseWriter, r *http.Request) error {
...
+		tags  = ""
 	)
 
+	// currently querries only are used to pass tags. However, this
+	// is extensible for future fields as well.
+	for _, s := range q["tag"] {
+		tags += s + ","
+	}
+	tags = strings.TrimSuffix(tags, ",")
+
```
activeshadow, line 45: I would suggest moving this down below the `if !role.Allowed` check just to avoid unnecessary work if the user can't create a workflow anyway.

Also, you can condense this down to a single line and get the same result.
```
tags := strings.Join(q["tag"], ",")
```
If `q` does not have a `tag` key, an empty slice will be returned. Doing a `strings.Join` on an empty slice simply results in an empty string.

src/go/web/workflow.go
```diff
@@ -114,7 +123,10 @@ func ApplyWorkflow(w http.ResponseWriter, r *http.Request) error {
 
 		defer cache.UnlockExperiment(expName)
 
-		annotations := map[string]string{"phenix.workflow/branch": scope}
+		annotations := map[string]string{
+			"phenix.workflow/branch": scope,
+			"phenix.workflow/tags":   tags,
```
activeshadow, line 128: What if `tags` is empty?

activeshadow (changes requested):
@nblair2 I did an initial review and added some comments. Overall it looks good!

To fix the issue of tags not getting added to an existing experiment, we'll need to update the `experiment.WriteToStore` function in `types/experiment.go` to update the config metadata along with the config spec and status.

nblair2:
@activeshadow suggested changes made, thanks you!

I also updated `WriteToStore` in `types/experiment.go` so that metadata is updated too.

activeshadow (approved):

PR #140, opened 2023-07-20
fix: execute post start stage of apps even when in dry-run mode

activeshadow:


src/go/api/experiment/experiment.go
```diff
@@ -507,22 +506,26 @@ func Start(ctx context.Context, opts ...StartOption) error {
...
 
-				if err := mm.ClearNamespace(exp.Spec.ExperimentName()); err != nil {
-					errors = multierror.Append(errors, fmt.Errorf("killing experiment VMs: %w", err))
-				}
+			if err := app.ApplyApps(context.TODO(), exp, app.Stage(app.ACTIONCLEANUP), app.DryRun(o.dryrun)); err != nil {
+				errors = multierror.Append(errors, fmt.Errorf("cleaning up app experiments: %w", err))
+			}
 
-				return errors
+			if err := mm.ClearNamespace(exp.Spec.ExperimentName()); err != nil {
```
glattercj, line 518: Why are we clearing the namespace if it's a dry run?

activeshadow: This line get executed regardless of whether or not this is a dry run. Clearing a namespace that doesn't exist shouldn't cause any errors.

glattercj (approved):

PR #144, opened 2023-07-31
Util command to print permissions and roles table

jacdavi:
This adds a command that prints a table showing all permission checks in phenix and whether each role can perform them.

Admittedly it's a bit janky. Wasn't sure of a good way to get all existing checks other than greping the source code. If there's a better way to do that, let me know.

Output defaults to CSV, but can be markdown with `--pretty`.

Example output of `./bin/phenix util role-table ../.. -p`:

|                      |        | DISABLED | EXPERIMENT ADMIN | EXPERIMENT USER | EXPERIMENT VIEWER | GLOBAL ADMIN | GLOBAL VIEWER | VM ADMIN | VM VIEWER |
|----------------------|--------|----------|------------------|-----------------|-------------------|--------------|---------------|----------|-----------|
| applications         | list   |          |                  |                 |                   | Y            | Y             |          |           |
| configs              | create |          |                  |                 |                   | Y            |               |          |           |
| configs              | delete |          |                  |                 |                   | Y            |               |          |           |
| configs              | get    |          |                  |                 |                   | Y            | Y             |          |           |
| configs              | list   |          |                  |                 |                   | Y            | Y             |          |           |
| configs              | update |          |                  |                 |                   | Y            |               |          |           |
| disks                | list   |          | Y                |                 |                   | Y            | Y             |          |           |
| exp/captureSubnet    | create |          |                  |                 |                   | Y            |               |          |           |
| experiments          | create |          |                  |                 |                   | Y            |               |          |           |
| experiments          | delete |          |                  |                 |                   | Y            |               |          |           |
| experiments          | get    |          | Y                | Y               | Y                 | Y            | Y             | Y        |           |
| experiments          | list   |          | Y                | Y               | Y                 | Y            | Y             | Y        |           |
| experiments          | patch  |          |                  |                 |                   | Y            |               |          |           |
| experiments          | update |          | Y                |                 |                   | Y            |               |          |           |
| experiments/apps     | get    |          | Y                | Y               | Y                 | Y            | Y             | Y        |           |
| experiments/captures | list   |          | Y                | Y               | Y                 | Y            | Y             | Y        |           |
| experiments/files    | get    |          | Y                | Y               | Y                 | Y            | Y             | Y        |           |
...


activeshadow:
@jacdavi this is cool and should be very useful. I can already see that I need to clean up some policy names. 😂

Besides using regex, the only other way I can think of to do it is using built in Go packages for finding comments and commenting each use of what you're searching for. But then that depends on it actually getting commented and I'm not sure how much value it adds doing it that way.

I'll take a closer look at this ASAP.

activeshadow:
@jacdavi since this utility command requires access to the source code, what if we instead created a little utility executable that does what you're doing here to generate both the CSV and markdown versions when phēnix is built, and then package those individual files in the resulting executable (as binary data) to be displayed when this command is run? The utility executable could be triggered via `//go:generate` at build time.

Thoughts?

activeshadow:
We should think about adding this to the UI as well... show a table of what the currently logged in user can do.

jacdavi:
Alright took a first stab at generating at build time. Seems to work fine, but I've never used `go:generate` before so I might've missed some conventions. Feel free to let me know if stuff needs to be changed or moved around.

activeshadow:
Thanks @jacdavi! I'll try to do another review and test of this code ASAP this week. Is there a certain date you need it merged by?

jacdavi:
Nope, no deadline. Just something I wanted to add :)

activeshadow:
Sorry @jacdavi I still need to get to this review...

jacdavi:
No problem! There's no rush on this one

activeshadow (commented):
@jacdavi can you take a quick look at the latest commit I added? Just some minor tweaks due to my OCD. :joy:

I'll then squash and merge!

jacdavi:
Looks good! I updated the generated struct to match the syntax you used as well

PR #147, opened 2023-08-11
feat: streaming netflow support for experiments

activeshadow:
* Enables creation of netflow capture for an experiment via API calls
* Adds `defaultBridge` setting to experiment schema to support the use of separate bridges per experiment
* Adds button to UI experiment page for starting and streaming netflow data for experiment

## Needs Review and Testing

* [x] Do more testing of default bridge setting.
  * Make sure existing experiments get updated correctly with a default bridge name of `phenix` when this version of phenix is used.
  * Make sure taps are created correctly if a bridge isn't specified in the config (should use default bridge for experiment if not set in tap config). This goes for the scorch tap component and the tap app.
* [x] Decide if netflow should always be streamed to the UI if a capture has been created, or if that should be something the user can also enable/disable.
  * Per a discussion w/ @glattercj, we've decided to leave this as-is for now (always streaming to UI via WebSockets when netflow is enabled).

## TODO

* [ ] ~Update the mirror app in phēnix-apps to set the default bridge correctly if not specified in the config.~
  * This will have to be done after this PR is merged in.

## Current State

Netflow is captured using the netflow capabilities within Open vSwitch. Since OVS only allows netflow to be configured and captured bridge-wide, the netflow feature in phēnix will fail if the default bridge for an experiment is set to the default of `phenix`, to avoid data leakage across experiments. Additionally, the creation or updating of an experiment will fail if a default bridge is used that another experiment is already configured with.

A new `defaultBridge` setting has been added to the experiment schema, and when set to something other than `phenix` it will allow experiment-wide netflow capture and will also automatically use GRE tunneling between the OVS bridges on multiple mesh nodes.

A new `--default-bridge` flag has been added to the `phenix experiment create` subcommand that will set the default bridge name when creating the experiment.

The experiment creation modal in the UI has a new option input for setting the default bridge name when creating experiments via the UI.

Bridge names must be 15 characters or less, as dictated by OVS (and Linux interface names).

When netflow is activated for an experiment, external applications can connect a websocket to `/api/v1/experiments/{name}/netflow/ws` in order to get streaming netflow data for the experiment.


activeshadow:
@glattercj any thoughts on my 2nd bullet in the main description above regarding whether or not enabling netflow for an experiment should always stream it to the UI by default? Right now, that's how it works.

Also, right now, if a separate process (via the API) enables netflow for an experiment, it can still be disabled via the UI (there's no ownership lock).

activeshadow:
Per a new "rule" I'm trying to enact (thanks @glattercj for the suggestion!), I'm going to leave this marked as draft until I have a corresponding PR in the phēnix docs repository.

PR #151, opened 2023-09-10
feat: Add a startup script delay to annotated domain controllers

glattercj:
Adds a wait loop in the windows startup script if this annotation is added:

```yaml
      annotations:
        phenix/domain-controller: "true"
```

The wait loop specifically waits for the `Get-ADDomain` command to return, meaning the machine has fully booted and active directory services have started. Without this delay, the DC machine may enter a reboot loop.


activeshadow:
@glattercj looks good to me. Is there a specific reason you have it marked as draft? Are you still doing some testing?

Also, can you maybe add something to the documentation about this? We probably need to add a section on annotations so we can document this and some of the other Windows-specific annotations we have.

PR #152, opened 2023-09-17
GetVMInfo:Disk info for VMs on remote nodes

eric-c-wood:
This fixes a Phenix VM list web page loading error that occurs when a disk image only exists on a remote node.  This is the case when doing a VM snapshot for a VM that exists on a remote node. The minimega `disk info` command which wraps `qemu-img` info is ran on the local node and does not appear to account for disk images on remote nodes.


activeshadow:
LGTM

@eric-c-wood I'm assuming you've tested this against an env that you witnessed the bug in?

eric-c-wood:
@activeshadow Yes, I did test the changes in the environment where I witnessed the bug.

PR #157, opened 2023-09-27
Changes to Docker build to improve dev workflows

GhostofGoes:
## Summary

The primary reason for this PR is to allow usage of non-GitHub repos to be used for `APPS_REPO` during the Docker image build. There are also a number of other enhancements to the Dockerfile and a handful of updates to the README (just some cleanup while I was in the area).

## Changes

- APPS_REPO must now include the base domain for the repo. By default, this is github.com. The change should have minimal impact on existing workflows. This change enables use of other Git hosts for phenix-apps.
- Add some labels to the image metadata.
This is a best practice that we use for several other projects.
- Disable pip update checks
- Disable pip caching
- Move scorch dependency installs before python installs, to improve caching (since these usually don't change, so docker buildkit can cache and not have to pull when rebuilding)
- Add ability to configure pip index via build arg.
This enables use of PyPI mirrors to speed up builds and dodge proxy errors.
- Remove --trusted-host arguments from pip. There's no reason to have these if the certificates are installed.
- Add a few comments explaining stages
- Update APPS_REPO in the docker GitHub Action workflow
- Add `.venv` to `.gitignore`
- Minor changes to `README.md`

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,22 +4,25 @@ Welcome to `phenix`!
 
 ## Building
 
-To build locally, you will need Golang v1.18, Node v14.2, Yarn 1.22, and Protoc
-3.14 installed. Once installed (if not already), simply run `make bin/phenix`.
+### Local build
+To build locally, you will need Golang v1.18, Node v14.2, Yarn 1.22, and Protoc 3.14 installed.
 
-If you don't want to install Golang and/or Node locally, you can also use Docker
-to build phenix (assuming you have Docker installed). Simply run
-`./docker-build.sh` and once built, the phenix binary will be available at
-`bin/phenix`. See `./docker-build.sh -h` for usage details.
-
-A Docker image is also hosted in this repo under Packages and can be pulled via:
-
+Once installed (if not already), simply run:
+```bash
+make bin/phenix
 ```
-$> docker pull ghcr.io/sandialabs/sceptre-phenix/phenix:main
-```
+
+If you don't want to install Golang and/or Node locally, you can also use Docker to build phenix (assuming you have Docker installed). Simply run `./hack/build/docker-build.sh` and once built, the phenix binary will be available at `bin/phenix`. Run `./hack/build/docker-build.sh -h` for usage details.
+
+### Docker build
+Docker is the recommended way to build and deploy phenix.
 
 The Docker image is updated automatically each time a commit is pushed to the
-`main` branch.
+`main` branch. To pull the latest image:
+
+```shell
+docker pull ghcr.io/sandialabs/sceptre-phenix/phenix:main
+```
 
 > **NOTE**: currently the `main` image available on GHCR defaults to
 > having UI authentication disabled. If you want to enable authentication,
@@ -51,7 +54,7 @@ These tags are stored as a string `key1=value1,key2=value2`.
 
 ### Add/Update Topology or Scenario Config
 
-```
+```bash
 curl -XPOST -H "Content-Type: application/x-yaml" \
   --data-binary @{/path/to/config/file.yml} \
   http://localhost:3000/api/v1/workflow/configs/{branch name}
@@ -59,7 +62,7 @@ curl -XPOST -H "Content-Type: application/x-yaml" \
 
 ### Apply phenix Workflow Config
 
-```
+```bash
 curl -XPOST -H "Content-Type: application/x-yaml" \
   --data-binary @{/path/to/config/file.yml} \
   http://localhost:3000/api/v1/workflow/apply/{branch name}[?tag=key1=value1&tag=key2=value2]
@@ -69,7 +72,7 @@ curl -XPOST -H "Content-Type: application/x-yaml" \
 
 Below is an example phenix workflow config file.
 
-```
+```yaml
 apiVersion: phenix.sandia.gov/v0
 kind: Workflow
 metadata: {}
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -4,22 +4,23 @@ Welcome to `phenix`!
...
 
-To build locally, you will need Golang v1.18, Node v14.2, Yarn 1.22, and Protoc
-3.14 installed. Once installed (if not already), simply run `make bin/phenix`.
+### Local build
+To build locally, you will need Golang v1.18, Node v14.2, Yarn 1.22, and Protoc 3.14 installed.
 
-If you don't want to install Golang and/or Node locally, you can also use Docker
-to build phenix (assuming you have Docker installed). Simply run
-`./docker-build.sh` and once built, the phenix binary will be available at
-`bin/phenix`. See `./docker-build.sh -h` for usage details.
```
activeshadow, line 13: This got moved to `hack/build/docker-build.sh` just in case you want to leave this in the README.

docker/Dockerfile
```diff
@@ -67,19 +70,33 @@ ARG PHENIX_TAG
...
 
 
+# ** Build the phenix image **
 FROM ubuntu:22.04
 
-ENV TZ=Etc/UTC
+# General image metadata
+# Reference: https://github.com/opencontainers/image-spec/blob/main/annotations.md
+ARG PHENIX_COMMIT
+LABEL org.opencontainers.image.authors="Sandia National Laboratories <emulytics@sandia.gov>" \
```
activeshadow, line 89: Since I'm one of the main authors, but not part of the emulytics@sandia.gov mailing list, can I get my name and email address added here as well?

docker/Dockerfile
```diff
@@ -92,8 +109,9 @@ RUN apt update \
   && rm -rf /var/lib/apt/lists/* \
   && rm -rf /var/cache/apt/archives/*
 
-ENV LANG   en_US.UTF-8
-ENV LC_ALL en_US.UTF-8
+ENV LANG="en_US.UTF-8" \
+  LC_ALL="en_US.UTF-8" \
+  PIP_DISABLE_PIP_VERSION_CHECK=1
```
activeshadow, line 114: I've not seen Dockerfile env variables split across lines like this... I'm assuming you've tested it?

docker/Dockerfile
```diff
@@ -108,30 +126,41 @@ RUN ["/bin/bash", "-c", "if [ -n $INSTALL_CERTS ]; then \
...
-RUN python3 -m pip install \
-  --trusted-host pypi.org \
-  --trusted-host files.pythonhosted.org \
-  "git+https://github.com/${APPS_REPO}.git@${APPS_BRANCH}#egg=phenix-apps&subdirectory=src/python"
-
 # needed to build Kali images with phenix
 RUN wget -O kali.deb https://http.kali.org/kali/pool/main/k/kali-archive-keyring/kali-archive-keyring_2022.1_all.deb \
-	&& dpkg -i kali.deb && rm kali.deb
+	&& dpkg -i kali.deb \
+  && rm kali.deb
```
activeshadow, line 132: If you're going to split this, can you untab the first line so they're both indented the same?

docker/Dockerfile
```diff
@@ -108,30 +126,41 @@ RUN ["/bin/bash", "-c", "if [ -n $INSTALL_CERTS ]; then \
...
-  && tar -xzf glow.tgz glow && mv glow /usr/local/bin/glow && rm glow.tgz
+  && tar -xzf glow.tgz glow \
+  && mv glow /usr/local/bin/glow \
+  && rm glow.tgz
 
 # used by scorch
 RUN wget -O filebeat.deb https://artifacts.elastic.co/downloads/beats/filebeat/filebeat-7.17.5-amd64.deb \
-	&& dpkg -i filebeat.deb && rm filebeat.deb
+	&& dpkg -i filebeat.deb \
+  && rm filebeat.deb
```
activeshadow, line 143: If you're going to split this, can you untab the first line so they're both indented the same?

docker/Dockerfile
```diff
@@ -108,30 +126,41 @@ RUN ["/bin/bash", "-c", "if [ -n $INSTALL_CERTS ]; then \
...
+# Allow the use of PyPI mirrors via build args
+ARG PIP_INDEX="https://pypi.org/"
+ARG PIP_INDEX_URL="https://pypi.org/simple"
+
+# Do this before installing phenix-apps so minimega package is latest version
+COPY --from=minimega /opt/minimega/lib /tmp/minimega
+RUN python3 -m pip install --no-cache-dir /tmp/minimega
+
+# Allow installing from a fork of sceptre-phenix-apps that could
+# exist on GitHub or elsewhere.
```
activeshadow, line 154: Maybe add this same comment above line 73 as well?

activeshadow (changes requested):
@GhostofGoes overall this looks good, though I have not tried building an image with it yet. I just had a few suggestions.

docker/Dockerfile
GhostofGoes: Yes, these changes were tested, and I use multi-line ENV split in a number of other projects. Essentially it bundles all the variables into single image layer, instead of a separate layer for each variable.

docker/Dockerfile
GhostofGoes: Heh, two tab characters snuck in somehow. Fixed.

docker/Dockerfile
GhostofGoes: Definitely. I added your name with the activeshadow email from your git commits. If there's another you'd prefer, let me know.

README.md
GhostofGoes: Totally missed that this got moved and not deleted. Updated README to point to it.

GhostofGoes:
@activeshadow Pushed changes to address review comments. These changes have been tested.

Issue #158, opened 2023-10-06
[UI-Bug] Cannot Create Experiment Without a Scenario


eric-c-wood:
In the recent change referenced below `this.createModal.scenarios[this.createModal.scenario]` will be undefined if an experiment does not contain a scenario.

https://github.com/sandialabs/sceptre-phenix/blob/12192c13f8391ac69e07364286ed50ebbd4673cf/src/js/src/components/Experiments.vue#L572

PR #159, opened 2023-10-09
fix(ui): allow experiment to be created without a scenario

causand22:
Adding a null check before creating disabledApps list. Create window works with and without a scenario selected. 

@activeshadow Not sure how to link this pull request with the issue -- this will resolve issue #158


eric-c-wood:
@activeshadow yes, this patch fixes issue #158.

PR #162, opened 2023-11-22
Disable extended clipboard in noVNC

jacdavi:
This is a minor update that coincides with a minimega PR adding copy/paste support: https://github.com/sandia-minimega/minimega/pull/1520

This disables noVNC from using the [extended clipboard pseudo encoding](https://github.com/rfbproto/rfbproto/blob/master/rfbproto.rst#extended-clipboard-pseudo-encoding). Minimega uses a passive shim to read RFB messages, and can't support the extended clipboard.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -30,7 +30,7 @@ jobs:
           context: .
           file: docker/Dockerfile
           build-args: |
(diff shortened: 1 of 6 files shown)


activeshadow:
LGTM

PR #164, opened 2024-01-18
feat: new deploy mode option for CLI and UI

glattercj:
- Remove `runLocal` option from experiment spec
- Replace with `deploy-mode` CLI and UI option for `experiment start...`.
- Must be one of `all | no-headnode | only-headnode`, with default=no-headnode.
- The effect of this option is to modify the minimega script `ns hosts` list for the experiment namespace
  * `all` - adds `localhost` to the namespace ns hosts list
  * `no-headnode` - default behavior; no changes to minimega script; uses mesh nodes as ns hosts list
  * `only-headnode` - removes all mesh nodes from ns hosts list and adds `localhost` only

### CLI example:
```bash
phenix exp start --deploy-mode=all foobar
```

### UI example:
```bash
phenix ui --unix-socket-endpoint=/tmp/phenix.sock --deploy-mode=only-headnode
```


activeshadow (approved):
@glattercj I've updated some of the code to achieve the following:

* Move the `--unix-socket-endpoint` option for the `ui` subcommand to now be a root `--unix-socket` option that is used by the UI to run a HTTP server at the path provided and by the CLI to attempt to connect to the HTTP server at the path provided. Previously, the UI would only start a HTTP server on the Unix socket if set explicitly. Now it will do so by default since a default path is provided.
* Moved the `--deploy-mode` option from the `ui` subcommand to be a root option that is used by either the UI server or any of the CLI subcommands. Most notably, the `exp create` subcommand. This deviates from what you had done in that it's used when creating an experiment instead of when starting/restarting one.
* If the CLI is able to connect to the UI server to get the `deploy-mode` option set when the UI was started, it will use that as the default deploy mode for any CLI subcommands unless the root `--deploy-mode` option is passed to the CLI command.
* In any case, if the `--deploy-mode` is not provided to the UI server and is not provided to the CLI, it will default to the traditional `no-headnode` mode.
* Added an option for setting the deploy mode to the UI modal for creating an experiment. If left unset, it will default to whatever the default mode is for the UI.
* Added an option for setting the deploy mode in the git workflow. The key is `deployMode` and is at the same level as `topology`, `scenario`, etc. If left unset, it will default to whatever the default mode is for the UI.

Thoughts?

activeshadow:
Please do not merge until I can rebase and squash these commits and update the commit message.

glattercj:
> @glattercj I've updated some of the code to achieve the following:
> 
> * Move the `--unix-socket-endpoint` option for the `ui` subcommand to now be a root `--unix-socket` option that is used by the UI to run a HTTP server at the path provided and by the CLI to attempt to connect to the HTTP server at the path provided. Previously, the UI would only start a HTTP server on the Unix socket if set explicitly. Now it will do so by default since a default path is provided.
> * Moved the `--deploy-mode` option from the `ui` subcommand to be a root option that is used by either the UI server or any of the CLI subcommands. Most notably, the `exp create` subcommand. This deviates from what you had done in that it's used when creating an experiment instead of when starting/restarting one.
> * If the CLI is able to connect to the UI server to get the `deploy-mode` option set when the UI was started, it will use that as the default deploy mode for any CLI subcommands unless the root `--deploy-mode` option is passed to the CLI command.
> * In any case, if the `--deploy-mode` is not provided to the UI server and is not provided to the CLI, it will default to the traditional `no-headnode` mode.
> * Added an option for setting the deploy mode to the UI modal for creating an experiment. If left unset, it will default to whatever the default mode is for the UI.
> * Added an option for setting the deploy mode in the git workflow. The key is `deployMode` and is at the same level as `topology`, `scenario`, etc. If left unset, it will default to whatever the default mode is for the UI.
> 
> Thoughts?

LGTM

PR #169, opened 2024-02-06
fix: provide options for VM CPU cores, threads, and sockets

### LINKED ISSUE #167: Unable to create guests with more than two cores
In relation to my issue here: <https://github.com/sandia-minimega/minimega/issues/1527>
I am currently not able to create guests with more that two cpu cores. That is true for both Windows and Linux OS guests. It would be helpful to specify at least socket count and core count. These information can be passed to minimega.

activeshadow:
Within the hardware section of the node configuration, at the same level as `vcpus`, you can now also provide `cores`, `threads`, and `sockets` settings.

fixes #167

### DIFF
--- a/src/go/tmpl/templates/minimega_script.tmpl
+++ b/src/go/tmpl/templates/minimega_script.tmpl
@@ -42,6 +42,9 @@ clear vm config
 vm config schedule {{ index $.Schedules .General.Hostname }}
         {{- end }}
 vm config vcpus {{ .Hardware.VCPU }}
+vm config cores {{ .Hardware.Core }}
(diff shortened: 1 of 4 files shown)


activeshadow:
@Narratot please test this when you can. I have not tested it at all.

Narratot:
@activeshadow I will check that, as soon as we start building larger systems. ATM I am using the changes, which I did within https://github.com/sandia-minimega/minimega/pull/1528 (I did split between specification of VCPUs and Cores + Sockets)

activeshadow:
Closing this in favor of #178.

PR #170, opened 2024-02-06
fix: avoid duplicates when creating new image from existing image

### LINKED ISSUE #168: image create-from adds duplicate scripts
When using the `phenix image create-from` command, the standard phenix scripts (`POSTBUILD_APT_CLEANUP`, `POSTBUILD_NO_ROOT_PASSWD`, `POSTBUILD_PHENIX_HOSTNAME`, `POSTBUILD_PHENIX_BASE`) are automatically added, even if the original config already has them.

Checking for the existence of these scripts in the original image config should be added.

activeshadow:
fixes #168


activeshadow:
@nblair2 see if this fixes the duplicates you're seeing. I have not tested.

PR #177, opened 2024-02-16
fix: avoid tunnel failures from expected errors

activeshadow:
When `cc tunnel` is run in a namespace that has multiple nodes, each node will try to create the tunnel and only one will succeed since the VM being used for the tunnel will only be on one node. The other nodes will report back an error, and this was causing phēnix to assume the tunnel creation had failed.

This commit updates the `cc tunnel` command to be prefixed with the node to execute the command on rather than having it sprayed to all nodes.


activeshadow:
I have tested this fix in an environment with multiple compute nodes.

PR #186, opened 2024-06-18
Validate JWT with UTC

glattercj:
Assume server is using UTC and match instead of using local time.


activeshadow:
@glattercj were you able to test this in your environment?

glattercj:
> @glattercj were you able to test this in your environment?

I tested it normally and it works...haven't been able to verify the exp mismatch yet.

PR #187, opened 2024-06-22
[UI] Fix infinite waiting when updating disks

eric-c-wood:
In the StoppedExperiment view, if there are no qcow images (i.e. .qcow2, .qc2) in the minimega files directory (usually `/phenix/images`), then the waiting indicator would spin indefinitely.  Likewise, in the RunningExperiment view, when using the `ChangeOpticalDisc` feature, the waiting indicator would spin indefinitely if there were no ISOs in the minimega files directory.  This patch fixes the issue with the infinite waiting indicator.

### DIFF
--- a/src/js/src/components/RunningExperiment.vue
+++ b/src/js/src/components/RunningExperiment.vue
@@ -1541,24 +1541,20 @@
     
       updateDisks (diskType="")  {
         this.disks = [];
-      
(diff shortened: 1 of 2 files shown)


activeshadow (approved):
LGTM, thanks @eric-c-wood!

PR #188, opened 2024-07-05
fix: remove colons from filenames output by scorch

nblair2:
Currently, scorch archive and info files are created using the start time in RFC3339 format, which includes colons. This causes issues with some OSes, where colons in filenames are not expected (ex. running `tar xzfv scorch-run-1_2024-07-03T23:02:41Z.tgz` will throw an error on linux systems).

All Golang [time.Format](https://go.dev/src/time/format.go) pre-defined layouts that include times also include colons, so this fix uses a custom format `2006-01-02T15-04-05Z0700`.

Alternatively `strings.ReplaceAll(startTime.Format(time.RFC3339, ":", "=")` could be used.

### DIFF
--- a/src/go/api/scorch/app.go
+++ b/src/go/api/scorch/app.go
@@ -165,7 +165,7 @@ func (this *Scorch) Running(ctx context.Context, exp *types.Experiment) error {
 	}
 
 	if _, err := os.Stat(runDir); err == nil {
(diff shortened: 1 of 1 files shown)


GhostofGoes:
Thanks for fixing this! This was an issue I ran into doing some experiments a few months ago, it make it difficult to share experiment data with collaborators on Windows systems (and harder for me to process and share the data).

PR #191, opened 2024-08-02
Add optional --unix-socket-gid flag

glattercj:
Adds an optional (and hidden) flag for specifying a group ID to apply group ownership to the unix-socket. Also sets the permissions to be group-writable. Use case: phenix starts from within a docker container as the `root` user, but normal users wish to be able to use the unix-socket to communicate with phenix (without having to switch to `root`). This allows the users to be a part of the group specified in the flag and have write access to the socket.

Normal (`--unix-socket-gid=1000`):
Screenshot 2024-08-02 at 08 49 19

Error (`--unix-socket-gid=foobar`):
Screenshot 2024-08-04 at 21 27 37


activeshadow (approved):
LGTM

@glattercj I made a few updates to the commit message and I'm in the process of testing and merging now.

PR #194, opened 2024-08-30
Various Fixes for Actions Taken on VMs

jacdavi:
Fixes a few bugs with the actions you can take on a booted VM

1. Fixes an error showing in UI after successful memory snapshot
2. Fixes miniccc not working in a VM after restoring a snapshot
  * This was caused by phenix not copying over the VM's UUID when creating the new VM with the snapshot disk. This resulted in the snapshot using the old UUID on the client, while the server expected a new value
4. Changes file mount for a VM to return available files along with any error
  * Previously we returned early if `os.Readdir` returned an error, but it returns the files read up until the error as well. This change returns those files. This fixes an issue in some directories (such as`/etc` on an Ubuntu host) being empty. Results should now be equivalent of mounting and running `ls` on command line
6. Removes reliance on `ps` for determining minimega's path
  * This requires the latest version of `minimega` where I added a new `args` command
8. Fix a few other web console errors


activeshadow:
Hey @jacdavi this sounds great! I'm glad you were able to figure out the miniccc snapshot issue. I've been trying to find time to look into that. 

I'll do my best (I know I say this a lot 😂) to get this and your other PRs reviewed and merged within the next week or so. I know you're already using them but they would for sure benefit the larger user base as well.

Thanks again!

activeshadow:
@jacdavi I've rebased this onto the latest main branch. I'll try to do some testing ASAP and get it merged in before main gets too far ahead again... :joy:

PR #195, opened 2024-09-04
feat(types): Add new helper funcs

glattercj:
Adds several helper functions for programmatically working with new  topology nodes.


GhostofGoes (approved):
LGTM

activeshadow:
@glattercj I'm assuming you've tested all of these additions with the Golang-based user app code you've been developing that needed them?

glattercj:
Correct. The app will likely change to be more general after our previous discussions, but these helper funcs should still be useful.

PR #202, opened 2025-03-06
Update the picture path when importing a builder XML file

davismax1:
Previously, the import from disk functionality only read in the XML data from the imported file and displayed the resulting topology to the user. The code modifications in this commit first get the current stencil path (which is where all the pictures used to create a visual topology are stored). Regex is then used to determine if the path currently used for pictures is different than the current stencil path. If they are different, the current stencil path replaces the appropriate portion of the old path for pictures. This prevents issues with the import from disk functionality that prior to these changes could import with broken pictures.


activeshadow:
Thanks @davismax1 for contributing! Do you think you could fix the indent spacing in the code you added please?

davismax1:
I definitely will! I'm making some further updates to the pull request but will make sure the comments are properly indented when I do so.

PR #209, opened 2025-04-02
feat: Add new Disks page to UI

jacdavi:
This adds a new UI tab for viewing and modifying disks. The UI Provides functionality for getting a detailed list of disks and allows for basic operations such as uploading, deleting, renaming, and copying.
It Also supports QEMU operations such as rebasing, snapshotting, and committing by wrapping minimega commands.. Everything should be ready to go after testing.

phenix_disk_table
phenix_disk_modal

Important notes:
* Previously there was code to handle disks across the mesh, but this has been removed. In a mesh, it is assumed that all disks are on the head node. minimega handles copying disks to other nodes at launch, and expects disks on the head.
* This depends on minimega [6b86c20](https://github.com/sandia-minimega/minimega/commit/6b86c2074b989afe87da7465846e880c77744b92) and [6b86c20](https://github.com/sandia-minimega/minimega/commit/6b86c2074b989afe87da7465846e880c77744b92). So the mm fork and docker arg will need to be updated before merging


jacdavi:
With this being merged, we will need to update the minimega version being used in the docker build: https://github.com/sandialabs/sceptre-phenix/blob/main/docker/Dockerfile#L4C17-L4C24

I can create a PR for that, but didn't include here originally because I wasn't sure of the cadence for updating that

Issue #211, opened 2025-04-04
Unable to build the image


LThai32422:
Hello,

I just started to learn Sceptre and followed the installation from this link:
https://sandialabs.github.io/sceptre-docs/quick-start.html#installation
and installed Sceptre, phenix, minimega.  After finished the installation i was able to startup Phenix GUI.

I tried the "Getting Started - Helloworld Experiment".  I was able to run step 1 (phenix image create -T /phenix/vmdb2/scripts/ubuntu --format qcow2 --release focal -c ubuntu).  But received an error at step 2 (phenix image build ubuntu -o /phenix -c -x).  This is the error:
.......
**Phenix 2025/04/03 20:52:36 [bdc19fd0-a3c3-4f1d-8b66-96a9f17f76c7] building image with vmdb2: exit status 1
Error: Unable to build the ubuntu image (search error logs for bdc19fd0-a3c3-4f1d-8b66-96a9f17f76c7)**

I spent a couple hours but could not figure out.  Can you please help.  Thanks.

glattercj:
Without more detail, I'm not sure what the error is, but it may be related to https://github.com/sandialabs/sceptre-phenix/issues/207. Try using an earlier commit as noted in that issue to see if it fixes things for you while a resolution is being worked on.

LThai32422:
Glattercj, thanks for your response.  This is the output.  Hope you can help me out.  Thanks.

root@sceptre:~# phenix image build ubuntu -o /phenix -c -x
INFO vmdb2 version 0.13.2+git starts
DEBUG sys.argv: ['vmdb2', '/phenix/ubuntu.vmdb', '--output', '/phenix/ubuntu', '--rootfs-tarball', '/phenix/ubuntu.tar', '-v', '--log', 'stderr']
DEBUG current working directory: /opt/phenix
DEBUG uid: 0
DEBUG effective uid: 0
DEBUG gid: 0
DEBUG effective gid: 0
DEBUG environment variables:
DEBUG environment: PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
DEBUG environment: HOSTNAME=sos
DEBUG environment: TERM=xterm
DEBUG environment: TZ=Etc/UTC
DEBUG environment: LANG=en_US.UTF-8
DEBUG environment: LC_ALL=en_US.UTF-8
DEBUG environment: PIP_DISABLE_PIP_VERSION_CHECK=1
[...]

PR #214, opened 2025-04-17
fix: upgrade vmdb2 to official upstream fork

### LINKED ISSUE #207: phenix image build broken since rebase to 24.04
I will troubleshoot this more when I have time (hopefully soon), but there is a bug with vmdb2 in the phenix docker image since the rebase to 24.04

Building any image results in an error like so:
```
ERROR: Command failed: debootstrap --variant minbase --include initramfs-tools,net-tools,isc-dhcp-client,openssh-server,init,iputils-ping,vim,less,netbase,curl,ifupdown,dbus,linux-image-generic,linux-headers-generic --components=main,restricted,universe,multiverse --verbose noble /tmp/tmpgxne8iq1 http://us.archive.ubuntu.com/ubuntu/
b"I: noble uses zstd compression, setting --extractor=ar option\nI: Retrieving InRelease \nI: Checking Release signature\nI: Valid Release signature (key id F6ECB3762474EDA9D21B7022871920D1991BC93C)\nI: Retrieving Packages \nI: Validating Packages \nE: Couldn't download http://us.archive.ubuntu.com/ubuntu/dists/noble/main/binary-amd64/Packages\n"
b''
```

The strange thing is that if you copy and run that same `debootstrap` command manually, it works fine. For now, you can use a phenix image at commit 01eea3edf0e5eb4b7a2fe956c7f09cd7d6119151 or older for image building.

glattercj:
# Upgrade vmdb2 to official upstream fork

## Description
This PR fixes the phenix image issues recently reported. It replaces the outdated vmdb2 binary with an upgraded version pointing to the official upstream fork on GitLab. Specifically, the old version at `github.com/glattercj/vmdb2` was archived. A fork of the actively-developed upstream https://gitlab.com/larswirzenius/vmdb2 was created at https://gitlab.com/glattercj/vmdb2. Custom plugins used in phenix were ported from the old version to the new fork. In this PR, the vmdb2 code in the Dockerfile and Containerfile were changed to use the new fork version of vmdb2 and the relevant `phenix image` code was updated to support it.

## Related Issue
Fixes #207 and #211.

## Type of Change
- [X] Bugfix
- [X] New feature

## Checklist
- [X] I have commented my code, particularly in hard-to-understand areas.
- [X] I have made corresponding changes to the documentation.
- [X] I have tested my code.

## Additional Notes
The reported issues were caused by a broken symlink in certain Ubuntu distributions as a result of systemd-resolved using a symlink for `/etc/resolv.conf`. That symlink points to `/run/systemd/resolve/stub-resolv.conf`, but during the phenix image build, `/run` was not mounted into the chroot environment, causing a broken symlink. The updated vmdb2 code fixes this and includes many other improvements and upgrades.

Major changes:
- Dockerfile/Containerfile clone new vmdb2 forked repo and use local run script instead of pulling a binary
- Modified packages in image constants based on testing
- Debootstrap "components" option is now a list instead of an options string and optionally specified similarly to the "packages" arg
- Removed `--very-very-verbose` option as it's not supported in the new vmdb2 plugin and the updated vmdb2 code has better logging (and the extra debootstrap logging didn't add much)
- Changed "--log" parameter in vmdb2 command to output to a file instead of stderr: `<image name>.log`
- Updated command line defaults for `phenix image` command: size=10G, release=jammy, components="", format=qcow2
- Updated vmdb2 template to match new vmdb2 code; this version includes mounting virtual filesystems for running systemd services inside chroot so for the majority of cases, you shouldn't need to build a custom `*.vmdb` script unless you need custom partitioning or e.g. UEFI boot
- Added prefix to default image configs to prevent overlap with custom images: `PHENIX-`
- Updated image constants to support debian/ubuntu/kali
- Removed deprecated flags: `--include-miniccc` & `--include-protonuke`
- Removed `miniccc.yml`, `protonuke.yml`, and `minirouter.yml` default images due to ^
- Updated default images to work with updated vmdb2 code

New features:
- Services can be ran inside phenix image builds (e.g. docker)
- Started services do NOT need to be manually stopped inside phenix image builds
- New vmdb2 code does a much better job cleaning up loop devices used during phenix image builds
- New vmdb2 code will use a new loop device if one is already in use (can run multiple phenix image builds simultaneously)
- Mounting virtual filesystems can be disabled by adding the `--no-virtuals` flag to the `phenix image create ...` command


docker/docker-compose.yml
```diff
@@ -23,10 +23,10 @@ services:
     image: phenix
     container_name: phenix
     privileged: true
+    pid: host
```
GhostofGoes, line 26: What is the difference between `pid: host` and mounting `/proc`? And should we make this change in other docker-compose files for phenix?

GhostofGoes (approved):
LGTM. Awesome work!

docker/docker-compose.yml
glattercj: This let's the phenix container directly access the host's PID tree, which is required for some of the advanced cleanup of image builds if they fail. With just the `/proc` mount, the phenix container doesn't have permissions to view the specific fuser information to be able to target and kill the "hung" processes preventing the unmounting of the chroot environment.

Issue #160, opened 2023-10-24
Upgrade Node.js from 14


jacdavi:
Looks like Node.js 14 is past EOL and the last update was 8 months ago. I've started getting warning when trying to install it.

Would be good to upgrade to a newer version of Node.js such as 20

https://nodejs.org/en/download/releases

Issue #122, opened 2023-04-06
[UI] Signin/Login Page Not Displayed


eric-c-wood:
If latest version of Phenix is built without using the Docker/Podman container workflows, the `VUE_APP_AUTH` variable in src/js/src/router.js is undefined resulting in a Javascript "too much recursion" error which results in the Signin/Login page not being displayed.   As a quick workaround, an environment file (i.e. .env, .env.local, etc.) is added to the src/js directory to explicitly set the  `VUE_APP_AUTH` environment variable.  It would be nice to either set the `VUE_APP_AUTH` variable to some default value or allow the `PHENIX_WEB_AUTH` to set the `VUE_APP_AUTH` variable when not using the container workflows.

activeshadow:
Hi @eric-c-wood, sorry I'm just now seeing this issue. I'll look into a solution for this ASAP. More to come.

Issue #11, opened 2021-04-16
[phenix image] Allow building image directly from vmdb config file


activeshadow:
---

Originally created by @glattercj on *Wed, 23 Dec 2020 17:19:43 GMT*

---
Add option to pass `phenix image build` an existing vmdb config file to build from, rather than having to first be in the image datastore. Alternatively, allow `phenix image create` the same option so an existing vmdb config can be added to the datastore and then built normally.

glattercj:
Why was this not fix again?

PR #236, opened 2025-07-18
feat: Vue 3 upgrade

causand22:
# Upgrade to Vue 3
Contributors: @jacdavi @causand22 
 
## Description
Updating the front end to use Vue 3 and newer versions of all tools. Quick overview:
- Node version updated to latest (22 at time of PR)
- Removing yarn for build process, using npm
- Vue 2 updated to Vue 3
  - store: Vuex -> Pinia
  - requests: vue resources -> axios
  - build tool: vue-cli -> vite
- buefy updated to v1.0 (vue 3 support) 
- Code organization, moved page base components to views/, separated some views into multiple components
- Separate CSS files 
- code formatted using prettier
- pages now load dynamically (making initial page loading quicker)
  - Config page has components load dynamically (editor dependencies) to make list load faster.

## Related Issue
- #160 
- #122 

## Type of Change
- [X] New feature
- [X] Documentation update

## Checklist
- [x] I have tested my code.

## Additional Notes
**KEY DEVELOPMENT CHANGE**
env arguments are different for vite. Instead of VUE_APP_AUTH, it's now VITE_AUTH. `.env` included with this PR.

Smaller changes:
- modifications to CSS
- Certain dependencies changed for lower page size
- added auto time out functionality (part of settings page)
  - changes made to go backend as well (added settings, one api route)
- various bugfixes  

Marking this PR as a draft as some features still need testing (scorch).


GhostofGoes:
Excellent! This is probably best reviewed by someone familiar with the Vue code, such as @activeshadow

glattercj:
@causand22 I might be doing something dumb, but I pulled

`docker pull ghcr.io/causand22/sceptre-phenix/phenix-jit-ui:feat-vue3`

and am getting this on startup of the UI:

image

jacdavi:
> > @causand22 I might be doing something dumb, but I pulled
> > `docker pull ghcr.io/causand22/sceptre-phenix/phenix-jit-ui:feat-vue3`
> > and am getting this on startup of the UI:
> > image
> 
> That's very likely on our end. We've never used the JIT image, so I hadn't looked at that more than making sure the image builds

@glattercj should be good now!

activeshadow (commented):
I have not reviewed yet, so I'm not approving or requesting changes at this point. What I will request is that we squash this down to a (much) smaller number of commits before merging. :joy:

I think I worry about the commit history a little too much...

cmulk:
@jacdavi  sorry we never reviewed this and a lot has happened since then so now there are lots of conflicts. I saw your note on #316, would you be willing to revisit this and resolve conflicts (and I guess upgrade to node 24 or 26 now!). If so I promise to not let it go stale this time

nblair2 (commented):
> 🤖 **AI-generated review** (Claude Code). Posted on behalf of @nblair2. Treat findings as suggestions to verify, not authoritative — file:line references were checked against the branch but please confirm before acting.

## Overview

A large, well-motivated modernization (124 files, +21k/−26k): Vue 2→3, Vuex→Pinia, vue-resource→axios, vue-cli→Vite, yarn→npm, Buefy 1.0, Node 22, plus a new server-backed idle-timeout/auto-logout feature and a `components/` → `views/` reorg. The direction is right and the reorg is sensible. However, **the migration is incomplete in spots** — there's a cluster of Vue-2-isms that break at runtime, concentrated in the Scorch views (which the PR description itself flags as "still need testing"). Reads as legitimately still-draft.

Severities: **🔴 runtime breakage**, **🟠 security/deploy**, **🟡 inconsistency**, **⚪ cleanup**.

---

## 🔴 Must-fix: runtime breakage

Each of these throws or silently no-ops at runtime.

**1. `.sync` modifier (removed in Vue 3) — 13 sites.** Silent no-op, so two-way binding is lost (table pagination, loading overlays, modals won't update).
- `:current-page.sync` → `v-model:current-page`: `Hosts.vue:14`, `Scorch.vue:22`, `Users.vue:210`, `Disks.vue:252`, `Experiments.vue:195`, `experiment/StoppedExperiment.vue:237,568`, `experiment/RunningExperiment.vue:1285`
- `:active.sync` → `v-model`: `Scorch.vue:118`, `ScorchRuns.vue:20,23,58`, `experiment/VMtilesView.vue:119`
- The same PR uses the correct `v-model`/`v-model:checked-rows` form elsewhere (e.g. `Hosts.vue:112`, `ConfigsList.vue:162`), so these are misses, not intent.

**2. `this.$set` (removed in Vue 3) — 5 active sites → `TypeError`.** `Scorch.vue:413`, `ScorchRuns.vue:236,336,344,352,365`. Vue 3 proxy reactivity makes these unnecessary — plain assignment works.

**3. `resp.body` (vue-resource API) used with axios — 10 sites → reads `undefined`.** axios puts the payload in `resp.data`. `Scorch.vue:427,428,532`, `ScorchRuns.vue:193,195,197,199,201,202,233`. (118 other call-sites correctly use `.data`.)

**4. Proxy-auth login path is broken.** `router.js:218` calls `store.commit('LOGIN')` — Pinia has no `.commit` (TypeError), and a redundant `store.login(...)` runs right after. The `.catch` at `router.js:222` uses `err.body.trim()` (vue-resource shape; axios uses `err.response.data`). Anyone running `VITE_AUTH=proxy` hits this on first navigation. There's also a dead commented-out `Vue.http` block (`router.js:224-240`) to remove.

**5. Leftover `this.errorNotification(err)` — 5 sites → crash in the error path.** The old global `Vue.mixin` was dropped, but these weren't converted to `useErrorNotification`: `components/configs/ConfigsList.vue:535`, `views/Disks.vue:451` (bare `errorNotification(err)` → ReferenceError), `views/ProxySignUp.vue:74`, `views/Settings.vue:132`, `views/StateOfHealth.vue:1375`. These fire exactly when the backend returns an error, so a real error gets masked by a JS exception.

**Theme:** items 1–3 cluster in `Scorch.vue`/`ScorchRuns.vue` — those two files are essentially unconverted, matching the PR's "scorch still needs testing" note.

**Lower-tier Vue-3 misses:** `<template slot="empty">` → `#empty` (`Scorch.vue:27`, `VMtilesView.vue:32`, slot silently not rendered); `.native` modifier removed (`SignIn.vue:44,60`, `VMMountBrowserModal.vue:16`, `VMLabelsModal.vue:44,59`, `RunningExperiment.vue:34,57`); `v-for` on `<template>` without `:key` (`Disks.vue:146`).

---

## 🟠 Security & backend

**6. Disk-resize validation regex is unanchored and over-restrictive** (`api/disk/disk.go:46`). `regexp.MustCompile(`[+-]?\d+[KMGTPE]`)` with `MatchString` is not anchored. Verified behavior:
- `"evil; disk resize x 100G"` → **passes** (substring match defeats the validation intent)
- `"100GB"` → passes, leaving trailing junk
- `"100g"` (lowercase) and `"512"` (bytes) → **rejected** — valid minimega sizes the UI doesn't prevent users from typing

Fix: anchor it — `^[+-]?\d+[KMGTPE]$` — and move the `MustCompile` to a package-level var (currently recompiled every call).

**7. JWT handling.** Token is stored in `localStorage` when "remember me" is set (same as before, not a regression), but the new `store.login` *also always* writes the token + role to `sessionStorage` regardless of `remember` (`store.js:30-42`), which is new. Separately, `websocket.ts:12` puts the JWT in the WS URL query string (`?token=...`), where proxies/servers routinely log it.

**8. Auto-logout is client-side only.** Expiry is checked in `router.beforeEach` (only on navigation); an idle user on a static route keeps a live token + open authenticated WS until they navigate. The timer does hit server `logout`, so it's not purely cosmetic — but real enforcement must be server-side JWT expiry (please confirm that's present). The new `GET /api/v1/settings/timeout` route (`server.go:266`) sits on the authenticated `api` subrouter alongside `/settings/password`, so its auth posture is fine, and it only exposes non-sensitive config.

**9. Go nits in the timeout feature.** `handlers.go:GetTimeoutSettings` logs `"Getting password settings:"` (copy-paste) and uses an inconsistent `plog.Error` signature (missing `plog.TypeSystem` on the first call). `timeout.go:UpdateTimeoutSettings` reports `"Timeout.Delay"` for both `TimeoutMin` and `WarningMin` errors (copy-paste).

---

## 🟠 Build / deploy

**10. `.env` — no secrets.** It's just `VITE_AUTH=enabled` plus comments; `.env.local` is gitignored. Behavior note: this flips the *default* dev auth to enabled while the Dockerfiles still default `PHENIX_WEB_AUTH=disabled`.

**11. `podman/Containerfile:22` is broken / inconsistent** — `npx build-docs bundle openapi.yml ...` vs the working `docker/Dockerfile:38` `npx @redocly/cli build-docs openapi.yml ...`. Wrong binary and wrong subcommand; the Podman docs build will fail.

**12. `.vite/deps/` cache is committed and not gitignored** (`src/js/.vite/deps/{_metadata.json,package.json}`). Vite's dep-optimizer cache — add `.vite` to `src/js/.gitignore` and untrack it.

**13. JIT image fragility.** The JIT Dockerfile dropped build-time `yarn install`; `docker/jit/phenix-ui-entrypoint.sh:17` now runs `npm install` on every container start (not `npm ci`, so the committed `package-lock.json` isn't honored deterministically). Worse, both the install and build redirect with `&>` (truncate) to the *same* log file and `res=$?` only captures the build — so a failed `npm install` is silently clobbered and the build proceeds. Worth hardening (`npm ci`, separate logs, check the install exit code).

**14. Version drift.** `node:22` floating tag in `docker/Dockerfile`/`podman/Containerfile`/`docker-build.sh` vs pinned `22.17.0` in the JIT image; `hack/build/docker-build.sh:80` still pins Go `1.20.14` while the images use `1.24.1`. Consider an `engines.node` floor in `package.json`. Dead deps: `sass-loader` (webpack-only) and `@types/lodash` (imports use `lodash-es` → should be `@types/lodash-es`).

Lockfile swap (yarn.lock deleted, package-lock added) is clean, and no functional `yarn`/`VUE_APP_*` references remain — good.

---

## 🟡 Inconsistencies

- **`App.vue:20-26` uses the bare `axios` import** (no `baseURL`, no auth interceptor) instead of the configured `axiosInstance` everyone else uses — works only because `/features` is unauthenticated.
- **Error handling is split three ways:** `useErrorNotification`, the (broken) `this.errorNotification`, and hand-rolled `$buefy.toast` branching in `SignIn.vue`.
- **`errorNotif.js:34`** calls `error.response.data.toLowerCase()` — throws when the 401 body is JSON (an object), defeating the auto-logout-on-invalid-token path; also a dead vue-resource `error.body` branch above it.
- **`websocket.ts`:** `send()` immediately after `connectWebsocket()` while the socket is still `CONNECTING` (`:75-84`, throws `InvalidStateError`); unbounded `2 ** numFailedConnects` backoff with no cap/jitter (`:42-48`); reconnect `setTimeout` handle never stored, so a logout can race a pending reconnect; `globalWsMessageHandler` dereferences `msg.resource.type` with no guard (`:108`).
- **`timeout.js:21`** throttle math is inverted (`this.time_set - now >= 1h` with `time_set` init to `0` is always a large negative → never throttles) and the `data.enabled || …` guard makes `fetchAndStart()` early-return permanently once enabled, so it never re-arms.
- **`formattingMixin.formatStringify`** now appends `.toLowerCase()` — a silent behavior change for tag/list display vs the old `stringify` filter.
- **`changePaginate()` source drift:** `Hosts.vue` reads the username from the Pinia store; `Users.vue`/`Experiments.vue` read `localStorage.getItem('user')` for the same key.

(Mixing Options API with a `<script setup>` import-only block is *not* a bug — setup imports hoist to module scope — but it's two paradigms per file.)

---

## ⚪ Simplification opportunities

- **A `useTable()` composable** would collapse the pagination `table` object (8 files), `changePaginate()` (7 files), and `filterString`/filtered-computed search logic (Disks/Experiments/Scorch/…) into one place — and let the `.sync` fixes happen once.
- **Remove dead code:** the commented `Vue.http` block in `router.js`, the `error.body` branch in `errorNotif.js`, the commented `visualizer()` scaffolding + its devDep in `vite.config.js`, and the commented NVM block in `docker-build.sh`.
- **`websocket.ts`** would be clearer as three direct `globalWs.onopen/onerror/onclose =` assignments rather than the string-keyed `forEach`/branch loop — which is where the reconnect bugs hide.

---

## Recommendation

Blocking items: the runtime-breakage cluster (#1–5, especially the Scorch views and proxy-auth path) and the unanchored disk-resize regex (#6). The build issues (#11–13) should be fixed too — the Podman build is broken and the `.vite` cache shouldn't be tracked. The 🟡/⚪ sections are non-blocking polish. The overall migration is solid and worth landing once the unconverted Scorch/proxy paths get the same treatment as the rest.

jacdavi:
> @jacdavi sorry we never reviewed this and a lot has happened since then so now there are lots of conflicts. I saw your note on #316, would you be willing to revisit this and resolve conflicts (and I guess upgrade to node 24 or 26 now!). If so I promise to not let it go stale this time

I probably phrased things poorly in #316 ,but unfortunately both @causand22 and I ran out of funding for emulytics work shortly after this PR was made.  Given the number of things claude identified, it would probably require someone else to bring it to the finish line, though hopefully it's 80% there already :)

nblair2:
> I probably phrased things poorly in #316 ,but unfortunately both @causand22 and I ran out of funding for emulytics work shortly after this PR was made. Given the number of things claude identified, it would probably require someone else to bring it to the finish line, though hopefully it's 80% there already :)

@jacdavi - checkout #320. Its all your work + some AI help to rebase and address the AI review above (Its AIs, all the way down). Happy to burn some tokens if it gets this across the finish line. I can do some light testing but obviously for something of this size more eyes are needed.

cmulk:
good news! just merged #320

PR #237, opened 2025-07-19
feat: SoH startupDelay, fix: SoH empty address, ns path

nblair2:
# feat: SoH startupDelay, fix: SoH empty address, ns path

## Description

One new feature and several fixes for SoH:

* **feat: `startupDelay`** (I am open to name changes) is a parameter that can be specified in order to delay SoH checks from running immediately at experiment start (or whenever the SoH app is reached). In every case that I have used SoH, I have found that it's first run in the experiment PostStart phase _always_ fails. This is due to various things, but basically "the experiment is still starting up, not all checks are going to work, yet".  This leads to a workflow of: start an experiment, wait for SoH to run in the PostStart phase (and fail), then manually run again either through the UI or CLI in order to get a "good" SoH check. Adding this delay allows slower configurations and processes to finish before SoH actually kicks off its checks, and when tunned can ensure that the first SoH run gives a "healthy" reading. This **will block** all other apps from completing, but SoH already blocks, and right now it is blocking for no reason because it always fails. This delay is only added to the PostStart phase, and does not affect later SoH runs in the Running phase.

* **fix:** if no address is set for a node in the topo, SoH no longer checks for a address. Previously SoH would check for a badly formated string (like `/0`) and fail.

* **fix:** revert a mistake I made in dfd9475 that incorrectly moved a miniccc file send path. This was based on testing with an outdated version of miniccc

## Related Issue
N/A

## Type of Change
- [x] Bugfix
- [x] New feature
- [ ] Documentation update

## Checklist
- [x] I have made corresponding changes to the documentation. (https://github.com/sandialabs/sceptre-phenix-docs/pull/14)
- [x] I have tested my code.

## Additional Notes
Please provide any additional information or context for your pull request here.

### DIFF
--- a/src/go/api/soh/app.go
+++ b/src/go/api/soh/app.go
@@ -119,6 +119,8 @@ func (this *SOH) PreStart(ctx context.Context, exp *types.Experiment) error {
 }
 
 func (this *SOH) PostStart(ctx context.Context, exp *types.Experiment) error {
+	logger := plog.LoggerFromContext(ctx, plog.TypeSoh)
+
 	if err := this.decodeMetadata(exp); err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


glattercj (approved):
LGTM

PR #268, opened 2026-01-07
fix: minimega_node hostname validation

nblair2:
# fix: minimega_node hostname validation

## Description

Update the node spec for hostnames to disallow special characters in hostnames. Previously we allowed any hostname, which caused problems when implemented in VMs. For example, setting `hostname: my_host` in a phenix topology would work. However, when the `hostname.sh` startup script runs, it calls `hostnamectl`, which will set the pretty name to `my_host`, but the static name would be sanitized to `myhost`. Then, when this VM checks in with minimega, it would use the static name of `myhost`. When phenix tries to execute miniccc commands against a host by looking for `my_host`, it fail because minimega does not have such a host.

There are various sources of documentation on linux and [windows](https://learn.microsoft.com/en-us/troubleshoot/windows-server/active-directory/naming-conventions-for-computer-domain-site-ou#computer-names) hostnames, most of which also reference the DNS RFC. But since we are directly dependant on `hostnamectl`, the [man page](https://man7.org/linux/man-pages/man1/hostnamectl.1.html) provides clear direction on allowed characters:
>         The static and transient hostnames must each be either a
>         single DNS label (a string composed of 7-bit ASCII lower-case
>         characters and no spaces or dots, limited to the format
>         allowed for DNS domain name labels), or a sequence of such
>         labels separated by single dots that forms a valid DNS FQDN.
>         The hostname must be at most 64 characters, which is a Linux
>         limitation (DNS allows longer names).

## Related Issue
N/A

## Type of Change
- [X] Bugfix
- ~New feature~
- ~Documentation update~
- ~Other (please describe):~

## Checklist
- ~I have commented my code, particularly in hard-to-understand areas.~
- [x] I have made corresponding changes to the documentation. https://github.com/sandialabs/sceptre-phenix-docs/pull/28/files
- ~My changes generate no new warnings.~
- [x] I have tested my code.

## Additional Notes
Please provide any additional information or context for your pull request here.

### DIFF
--- a/src/go/types/version/v1/schema.go
+++ b/src/go/types/version/v1/schema.go
@@ -224,6 +224,8 @@ components:
             hostname:
               type: string
               minLength: 1
+              maxLength: 63
+              pattern: '^[a-zA-Z0-9](?:[a-zA-Z0-9-]*[a-zA-Z0-9])?$'
               example: ADServer
             description:
               type: string
--- a/src/go/types/version/v2/schema.go
+++ b/src/go/types/version/v2/schema.go
@@ -250,6 +250,9 @@ components:
           properties:
             hostname:
               type: string
+              minLength: 1
+              maxLength: 63
+              pattern: '^[a-zA-Z0-9](?:[a-zA-Z0-9-]*[a-zA-Z0-9])?$'
               example: ADServer
             description:
               type: string


nblair2:
Lastly, discovered that linux cang do 64 char hostnames but windows can only do ... 63?
image

GhostofGoes:
Also, as a general note to that might be worth putting in docs, the hostname field shouldn't be used for FQDN, there's a 'dns:' field for that.

GhostofGoes (approved):

PR #311, opened 2026-05-27
feat: implement 'phenix vm set' command

GhostofGoes:
# feat: implement 'phenix vm set' command

## Description

Implement the `phenix vm set` command. This command is currently present as a stub with no implementation. This implements it.

## Related Issue
N/A

## Type of Change
- [ ] Bugfix
- [x] New feature
- [ ] Documentation update

## Checklist
- [x] I have tested my code.

## Additional Notes

Will update documentation later this week.

### DIFF
--- a/src/go/api/vm/vm.go
+++ b/src/go/api/vm/vm.go
@@ -293,7 +293,7 @@ func Get(expName, vmName string) (*mm.VM, error) {
 	return vm, nil
 }
 
-func Update(opts ...UpdateOption) error { //nolint:funlen // complex logic
+func Update(opts ...UpdateOption) error { //nolint:funlen,gocyclo,cyclop  // complex logic
 	o := newUpdateOptions(opts...)
 
 	if o.exp == "" || o.vm == "" {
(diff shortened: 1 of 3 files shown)


cmulk (changes requested):
I think it would be helpful [here](https://github.com/sandialabs/sceptre-phenix/blob/0809f7a1436a1e5af9762b40bb3c0a1479ef0104/src/go/api/vm/vm.go#L326) to also add an error message if a user tries to set other things (like cpu) with the labels/interfaces when an experiment is running (e.g., `phenix vm set exp vm --cpu 4 -L role=web`) . Right now I think it just silently ignores the other flags

ex:
```go
if running {
    unsupported := o.cpu != 0 ||
        o.mem != 0 ||
        o.disk != "" ||
        o.partition != 0 ||
        o.dnb != nil ||
        o.host != nil ||
        o.snapshot != nil

    if unsupported {
        return errors.New("only interface connections and tags can be updated while experiment is running")
    }

    if o.iface == nil && o.tags == nil {
        return errors.New("only interface connections and tags can be updated while experiment is running")
    }

    ...
}
```

GhostofGoes:
> I think it would be helpful [here](https://github.com/sandialabs/sceptre-phenix/blob/0809f7a1436a1e5af9762b40bb3c0a1479ef0104/src/go/api/vm/vm.go#L326) to also add an error message if a user tries to set other things (like cpu) with the labels/interfaces when an experiment is running (e.g., `phenix vm set exp vm --cpu 4 -L role=web`) . Right now I think it just silently ignores the other flags
> 
> ex:
> 
> ```go
> if running {
>     unsupported := o.cpu != 0 ||
>         o.mem != 0 ||
>         o.disk != "" ||
>         o.partition != 0 ||
>         o.dnb != nil ||
>         o.host != nil ||
>         o.snapshot != nil
> 
>     if unsupported {
>         return errors.New("only interface connections and tags can be updated while experiment is running")
>     }
> 
>     if o.iface == nil && o.tags == nil {
>         return errors.New("only interface connections and tags can be updated while experiment is running")
>     }
> 
>     ...
> }
> ```

fixed

PR #334, opened 2026-07-25
fix(mm): add additional checks and safety

nblair2:
# fix(mm): add additional checks and safety

## Description
  * Read experiment scripts as `namespace <exp> read <file>`. The bare `namespace <exp>` line set minimega's process-global active namespace and never put it back, so every other client's unprefixed commands landed there. The cc and packet-capture scripts, which carried no namespace line at all, are now scoped too.
  * Give timed-out commands their own connection so a timeout can no longer tear down the shared one and truncate other in-flight commands. The deadline now bounds the whole exchange, not just dispatch.
  * Report a lost connection as an error; every mmcli helper read it as "no results".
  * Redial on any sticky connection error rather than two hardcoded strings. A restarted minimega reports "server disconnected" and wedged phenix until phenix itself was restarted.
  * IsHeadnode no longer calls every node the headnode when the mesh listing is empty, silently dropping the `mesh send <host>` prefix. Headnode() caches the last known good name.
  * Drain response channels in GetVMScreenshot and MeshShellResponse instead of returning early and wedging the connection lock.
  * Address C2 commands by the name minimega launched the VM under, and filter by UUID where known. The liveness check folds case but `cc filter name=` and `cc mount` do not, so a mis-cased name passed the check and then targeted zero clients.
  * Keep sentinel identity on errors crossing a response channel, so `errors.Is(err, ErrTimeout)` can be true. Requires go-multierror 1.1.1.
  * First tests for mmcli, driven by a fake minimega socket.

## Related Issue
Follow-up to #313 / [sceptre-phenix-apps#110](https://github.com/sandialabs/sceptre-phenix-apps/pull/110).
Pairs with [sceptre-phenix-apps#122](https://github.com/sandialabs/sceptre-phenix-apps/pull/122).
Pairs with [sandia-minimega/minimega#1618](https://github.com/sandia-minimega/minimega/pull/1618)

## Type of Change
- [x] Bugfix
- [ ] New feature
- [ ] Documentation update

## Checklist
- [X] I have commented my code, particularly in hard-to-understand areas.
- ~I have made corresponding changes to the documentation.~
- [x] I have tested my code.

## Additional Notes
AI-generated. Human reviewed and tested.

### DIFF
--- a/src/go/util/mm/minimega.go
+++ b/src/go/util/mm/minimega.go
@@ -30,8 +30,9 @@ var (
 	ErrScreenshotNotFound = errors.New("screenshot not found")
 )
 
-const vmInfoCmd = "vm info"
 const (
+	vmInfoCmd = "vm info"
+
 	c2ActiveCheckInterval  = 2 * time.Second
 	responseWaitInterval   = 1 * time.Second
 	responseRegexGroupUUID = 2
@@ -43,13 +44,23 @@ const (
 // instances of the Minimega struct.
 var ccMu sync.Mutex //nolint:gochecknoglobals // global lock
 
+// Cache of the last known good headnode name. The headnode does not change for
+// the life of the process.
+var headnode struct { //nolint:gochecknoglobals // process-lifetime cache
+	mu   sync.Mutex
+	name string
+}
+
 // Regular express to use for matching C2 response headers.
 var responseRegex = regexp.MustCompile(`(\d*)\/(.*)\/(stdout|stderr):`)
 
 type Minimega struct{}
 
-func (Minimega) ReadScriptFromFile(filename string) error {
-	cmd := mmcli.NewCommand()
+// ReadScriptFromFile runs `read` against ns. The namespace has to travel
+// with the command: minimega's `read` prefixes every line of the script
+// with the namespace that was active when it started.
+func (Minimega) ReadScriptFromFile(ns, filename string) error {
+	cmd := mmcli.NewNamespacedCommand(ns)
 	cmd.Command = "read " + filename
 
 	err := mmcli.ErrorResponse(mmcli.Run(cmd))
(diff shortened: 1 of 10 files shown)


src/go/util/mm/minimega.go
```diff
@@ -1058,19 +1108,30 @@ func (Minimega) IsC2ClientActive(opts ...C2Option) error {
...
-	o := NewC2Options(opts...)
-
 	exec := func(ns, vm, cmd string) (string, error) {
 		ccMu.Lock()
 		defer ccMu.Unlock()
 
 		c := mmcli.NewNamespacedCommand(ns)
-		c.Command = "cc filter name=" + vm
+
+		// Filter by UUID where we have one. `cc filter name=` falls through to
```
GhostofGoes, line 1127: Are there cases where a VM would have a name but not a UUID? I thought every VM had a UUID.

src/go/util/mm/minimega.go
```diff
@@ -1445,19 +1506,33 @@ func (Minimega) MeshShellResponse(host, command string) (string, error) {
...
 
+	var (
+		out   string
+		found bool
+	)
+
+	// Drain the full response channel -- returning early would abandon it and wedge the shared minimega connection.
 	for resps := range mmcli.Run(cmd) {
 		for _, resp := range resps.Resp {
+			if found {
```
GhostofGoes: Should error check happen before `found` check? If I'm reading correctly, if there are multiple command errors, only one will be logged.

nblair2: You're right. I guess the thinking was "log the first, quickly drain the rest". Let me check if maybe only one error could get into the channel or there is some valid reason not to log first.

Will change if appropriate.

src/go/util/mm/minimega.go
nblair2: You are correct.  `vmUUID` _should_ be set (returned by `c2Client(o)` in line 1116) unless the `skipActiveClientCheck` option is set. So we only fall through to VM name in 2 cases where that check is skipped:

* [`startup.go`](https://github.com/sandialabs/sceptre-phenix/blob/e21580fc86c340ee16e7afe1894391b866c1c528/src/go/app/startup.go#L318) - post start windows command - maybe should be set here
* [`web/mount.go`](https://github.com/sandialabs/sceptre-phenix/blob/e21580fc86c340ee16e7afe1894391b866c1c528/src/go/web/mount.go#L168) - API unmount (an existing mount) - I don't think we need to check if the c2 client is active, just make a best effort to unmount.

cmulk:
not ignoring this, just trying to get time to review in more detail

nblair2:
> not ignoring this, just trying to get time to review in more detail

No rush, the real money is in the minimega namespace changes.

cmulk:
> > not ignoring this, just trying to get time to review in more detail
> 
> No rush, the real money is in the minimega namespace changes.

does the mm PR need to be merged before this, or is this still compatible with the current minimega:master?

nblair2:
> > > not ignoring this, just trying to get time to review in more detail
> > 
> > 
> > No rush, the real money is in the minimega namespace changes.
> 
> does the mm PR need to be merged before this, or is this still compatible with the current minimega:master?

No, but I haven't tested this independently. Since there is no depends-on feature in GitHub (should we switch to gitlab (j/k)), I'll just leave as draft until then.

GhostofGoes:
> > > > not ignoring this, just trying to get time to review in more detail
> > > 
> > > 
> > > No rush, the real money is in the minimega namespace changes.
> > 
> > 
> > does the mm PR need to be merged before this, or is this still compatible with the current minimega:master?
> 
> No, but I haven't tested this independently. Since there is no depends-on feature in GitHub (should we switch to gitlab (j/k)), I'll just leave as draft until then.

Maybe the new 🥞 https://docs.github.com/en/pull-requests/how-tos/stacked-pull-requests

PR #347, opened 2026-08-01
fix(vrouter): preserve configured router hostnames

### LINKED ISSUE #335: [BUG] Router hostnames converted to lower
### Description
phenix normally sets a guest's hostname verbatim from the topology name:
* Linux via `hostnamectl set-hostname {{ . }}` (tmpl/templates/linux_hostname.tmpl)
* Windows via `Rename()`

Routers are the one exception: `Node.RouterName()` (types/version/v1/node.go) force-lowercases the name and rewrites `.` and `_` to `-`.

Two addressing mechanisms behave oppositely on case:
* phenix Go's `cmd.Filters` (used by `IsC2ClientActive`, `GetVMInfo`, `GetVMHost`, `GetVMState`) goes through minicli's `.filter`, which lowercases both sides: **case-insensitive**.
* `cc filter name=` falls through to an implicit tag filter compared against the VM name with a plain `!=` in `internal/ron/client.go` — case-sensitive. Same for `cc filter hostname=`, and `cc mount`/`clear`, `cc mount`/`cc exitcode` via FindVM's exact `==`.

### Steps to Reproduce
N/A

### Expected Behavior
Need to trace where assumptions are made and why

### Actual Behavior
Things get converted and maybe missed

### Screenshots
N/A

### Environment
N/A

### Additional Context
N/A

### Checklist
- [X] I have included no proprietary/sensitive information in my issue.

GhostofGoes:
## Description
Preserve router hostname casing so guest, topology, and C2 identifiers remain consistent.

Marking as draft so questions posted in #335 can be resolved.

## Related Issue
Closes #335.

## Type of Change
- [x] Bugfix
- [ ] New feature
- [ ] Documentation update

## Checklist
- [x] I have tested my code.

## Additional Notes
Untested.

### DIFF
--- a/src/go/types/version/v0/node.go
+++ b/src/go/types/version/v0/node.go
@@ -479,15 +479,7 @@ func (n Node) FileDeletions() string {
 }
 
 func (n Node) RouterName() string {
-	if !strings.EqualFold(n.TypeF, "router") {
-		return n.GeneralF.HostnameF
-	}
-
-	name := strings.ToLower(n.GeneralF.HostnameF)
-	name = strings.ReplaceAll(name, ".", "-")
-	name = strings.ReplaceAll(name, "_", "-")
-
-	return name
+	return n.GeneralF.HostnameF
 }
 
 func (h Hardware) DiskConfig(snapshot string) string {
--- a/src/go/types/version/v0/node_test.go
+++ b/src/go/types/version/v0/node_test.go
@@ -1,0 +1,16 @@
+package v0
+
+import "testing"
+
+func TestNodeRouterNamePreservesHostname(t *testing.T) {
+	const hostname = "Branch-Router-01"
+
+	node := Node{
+		TypeF:    "router",
+		GeneralF: &General{HostnameF: hostname},
+	}
+
+	if got := node.RouterName(); got != hostname {
+		t.Fatalf("RouterName() = %q, want %q", got, hostname)
+	}
+}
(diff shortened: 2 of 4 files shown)


nblair2:
The question - why are those changes made (vyos, vyatta, minirouter requirements)? If these are requirements they should be added to the schema, validated at configure time.

GhostofGoes:
Dug into this. `RouterName()` came in with the original Vyatta app in f7c9d358 (April 2020). No commit explains it, and it only ever fed `system host-name` in `vyatta.tmpl`. Minirouter never used it, and the startup app skips routers.

Vyatta and every VyOS release validate `host-name` against `[A-Za-z0-9][-.A-Za-z0-9]*[A-Za-z0-9]` (max 63) and never lowercase it. Lowercasing was never a requirement. Underscore is the only rejected character, and the schema pattern from 65f8e83e already rejects `_` and `.` when a config is created. Firewall nodes use the same template and were never rewritten either. I removed `RouterName()` and render the hostname as written.

That regex also needs at least 2 characters, and Windows DNS host names have the same minimum, so #429 raises the schema's hostname minimum from 1 to 2 (docs: sandialabs/sceptre-phenix-docs#70). sandialabs/sceptre-phenix-apps#148 replaces the apps test fixture that modeled the old rewrite.

One caveat: a config created with `--skip-validation` that has `_` in a router hostname will now have VyOS reject the host-name line instead of getting `-` in its place.

Issue #247, opened 2025-09-04
[BUG] Cyclic Versioning of phenix/types/versions/v*


nblair2:
### Description
Currently, the `phenix/types/version/v1` package [imports](https://github.com/sandialabs/sceptre-phenix/blob/46d4df4b596b5b818e32be93a02101b8d6c3c7bc/src/go/types/version/v1/experiment.go#L10) `phenix/types/version/v2`, which means `v2` cannot import anything from `v1` without creating an import cycle.

### Additional Context
I was adding a new `TopologySpec` to `v2` and wanted to reuse the existing `v1` `Node` struct, but that introduces a cyclic import. A few paths forward:

- Create an internal domain model and use adapters between each versioned spec and the domain model. More work up front, but clean and future‑proof.
- Keep versions consistent across all types (introduce `v2` equivalents for any reused `v1` types) to avoid cross‑version imports. Easier short term, but leads to duplication.
- Implement the changes in `v1` topology if they are backward‑compatible, deferring a proper refactor.

### Checklist
- [X] I have included no proprietary/sensitive information in

activeshadow:
By the way, I'm 100% positive I'm the one that imported v2 into v1. I'm thus the one that didn't implement this properly.

I was likely trying to avoid requiring an upgrade to the experiment spec version in yaml configs when I upgraded the scenario versioning (a long time ago).

I should have never put a forward looking import into a v1 package, so I likely made this harder to fix now (existing experiments will have to be updated to be v2 if we remove the v2 reference and create a new v2 experiment spec). We do have existing code to provide an auto upgrade path for specs, we'd just have to see if that approach would work for experiments (I think it would...).

activeshadow:
> Based on that pattern of imports, just changing the organization from `phenix/types/version/v1/scenario.go` (grouped by version number) to `phenix/types/version/scenario/v1.go` (grouped by type) might work - to keep things independent of version and allow experiment versions to grow at a different rate than scenario versions, for example.

This is definitely a good idea worth looking into to see if there would be any circular dependency requirements with this type of organization. Would also need to see how much of a refactor this would burden the rest of the code base with.

PR #348, opened 2026-08-01
feat(app): allow nodes to skip default apps

### LINKED ISSUE #156: Option to not apply default user apps
It is desired to have an option to skip the application of the default user apps.  The recently added `external` node attribute will allow the default user apps to be skipped but the `external` nodes will not be managed by Phenix/Minimega.   We need a method that allows the default user apps to be skipped for VMs that will still be managed by Phenix/Minimega.

GhostofGoes:
# feat(app): allow nodes to skip default apps

## Description
Add ability to skip default apps for a node using an annotation, `phenix/default-apps: false`. This applies to all lifecycle stages.

## Related Issue
Closes #156.

Documentation: https://github.com/sandialabs/sceptre-phenix-docs/pull/50

## Type of Change
- [ ] Bugfix
- [x] New feature
- [ ] Documentation update

## Checklist
- [x] I have tested my code.

## Additional Notes
Tested by adding the annotation to a VM, and verifying that the default apps don't run. Not perfect.

Test1, which has the annotation to disable default apps, and also has a injection (`1_example-injection.sh`) that test2 doesn't. 
Note the hostname is incorrect, and the 3 typical startup injections aren't present.
image

Test2, which doesn't have the annotation, but is otherwise almost identical to test1 in configuration:
image

### DIFF
--- a/src/go/app/startup.go
+++ b/src/go/app/startup.go
@@ -157,6 +157,13 @@ func (s Startup) PreStart(ctx context.Context, exp *types.Experiment) error {
 			continue
 		}
 
+		// Skip generating startup scripts/templates for this node if default
+		// apps are disabled, now that the checks/normalization above (which
+		// should always run) have been performed.
+		if !defaultAppsEnabled(node) {
+			continue
+		}
+
 		// Check to see if a scenario exists for this experiment and if it
 		// contains a "startup" app. If so, store it for later use
 		var startupApp ifaces.ScenarioApp
@@ -309,8 +316,10 @@ func (Startup) PostStart(ctx context.Context, exp *types.Experiment) error {
 			continue
 		}
 
-		if strings.EqualFold(node.Hardware().OSType(), "windows") {
-			// Windows 10 doesn't automatically run scripts in the startup folder
+		// Windows 10 doesn't automatically run scripts in the startup folder;
+		// only re-run it here if default apps (which set up that script) are
+		// enabled for this node.
+		if defaultAppsEnabled(node) && strings.EqualFold(node.Hardware().OSType(), "windows") {
 			if ver, ok := node.GetAnnotation("windows-version"); ok && (ver == "10" || ver == 10) {
 				_, err := mm.ExecC2Command(
 					mm.C2NS(exp.Metadata.Name),
@@ -326,6 +335,8 @@ func (Startup) PostStart(ctx context.Context, exp *types.Experiment) error {
 			}
 		}
 
+		// The autotunnel annotation is independent of default apps, so it is
+		// honored regardless of whether default apps are enabled for this node.
 		if annotation, ok := node.GetAnnotation("phenix/startup-autotunnel"); ok {
 			var tunnels []string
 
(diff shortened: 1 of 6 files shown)


src/go/app/startup.go
```diff
@@ -57,6 +57,10 @@ func (s Startup) PreStart(ctx context.Context, exp *types.Experiment) error {
 	}
 
 	for _, node := range exp.Spec.Topology().Nodes() {
+		if !defaultAppsEnabled(node) {
+			continue
+		}
+
```
nblair2: I would move the early-bailout to ~163. Right now you are skipping some checks:
* 64 - 120: duplicate IPs (even if we don't want the startup app to run, we still want to check these)
* 122: external - bails anyways
* 126 - 157: disk present (even if we don't want the startup app to run, we still want to check / normalize).

src/go/app/startup.go
```diff
@@ -305,7 +309,7 @@ func (s Startup) PreStart(ctx context.Context, exp *types.Experiment) error {
 
 func (Startup) PostStart(ctx context.Context, exp *types.Experiment) error {
 	for _, node := range exp.Spec.Topology().Nodes() {
-		if node.External() {
+		if node.External() || !defaultAppsEnabled(node) {
```
nblair2: Again, maybe too early? The windows 10 command should be skipped, but what about autotunnel? I guess that is a part of the startup app????

nblair2 (changes requested):
1. I don't like how we keep shoving functionality into annotations to avoid modifying the spec. Yes, the spec needs work (eg #247), but  annotations aren't validated so someone could probably break this with ` phenix/default-apps: "false"` (string not bool).
2. How does this play with `includes-phenix-startup` annotation (undocumented, I believe)
3. Future might be nice to refactor some of the apps because it looks like startup is doing a ton of lifting outside its scope (autotunnel, validate IPs (this one is my fault)).
4. Future might be nice to have by-app disabled rather than this sledgehammer.

GhostofGoes:
This made more sense as an annotation, since it's a "edge case" that is rarely used except for debugging or niche use cases, which I believe is what annotations were built for.

Honestly, this is really hacky. How I would prefer to architect it is have a list of apps that that run, then can filter that list as needed. Also move more stuff out of startup because it's becoming bloated like sceptre.py.

nblair2:
> This made more sense as an annotation, since it's a "edge case" that is rarely used except for debugging or niche use cases, which I believe is what annotations were built for.
> 
> Honestly, this is really hacky. How I would prefer to architect it is have a list of apps that that run, then can filter that list as needed. Also move more stuff out of startup because it's becoming bloated like sceptre.py.

Agree with all of the above - not trying to hold this up just noting things we can add to the backlog. At this point I have lost track of how many custom (often undocumented) things you can do with annotations. Its great up front to get the feature, but  you lose the ability to get nice yaml validation. So I am thinking 2 new issues in the backlog:

* (first) #247
* Then we can scrub annotations and move things back into the proper schema
* in parallel we can look at cleaning up startup app.

src/go/app/startup.go
cmulk: agreed that makes sense to me

GhostofGoes:
updated per review comments

PR #354, opened 2026-08-03
fix(cmd): use default root store when sudoing to root

### LINKED ISSUE #14: Use /etc/phenix/store.bdb by default when root, even if root was obtained via sudo
---

Originally created by @activeshadow on *Wed, 20 Jan 2021 18:06:02 GMT*

---
Right now, `phenix` checks to see if the `SUDO_USER` env variable is set, and if so it uses the store local to that user. However, `SUDO_USER` is set even when a user is currently root via `sudo su`.

The hack way to get around this is to run `su -` after running `sudo su`, which unsets the `SUDO_USER` env variable.

In the code, we should also check the `SUDO_COMMAND` env variable and only use the store local to the `SUDO_USER` if the command is for the phenix executable. When someone sudo's to root, the `SUDO_COMMAND` will be set to something like `/bin/su`.

GhostofGoes:
# fix(cmd): use default root store when sudoing to root

## Description
Use default store for root when running as root if the root user was activated using `sudo ...`, e.g. `sudo su`.

## Related Issue
Fixes #14

## Type of Change
- [x] Bugfix
- [ ] New feature
- [ ] Documentation update

## Checklist
~~- [ ] I have made corresponding changes to the documentation.~~
- [x] I have tested my code.

## Additional Notes
Tested with a phenix binary from a built container on a Ubuntu host.

### DIFF
--- a/src/go/cmd/root.go
+++ b/src/go/cmd/root.go
@@ -429,9 +429,13 @@ func getCurrentUserInfo() (string, string) {
 		sudo = os.Getenv("SUDO_USER")
 	)
 
-	// Only trust `SUDO_USER` env variable if we're currently running as root and,
-	// if set, use it to lookup the actual user that ran the sudo command.
-	if u.Uid == "0" && sudo != "" {
+	// `SUDO_USER` (and `SUDO_COMMAND`) remain set in the environment even after
+	// a user has escalated to a root shell (eg. via `sudo su` or `sudo -i`), so
+	// checking `SUDO_USER` alone isn't enough to determine whether phenix was
+	// invoked directly via sudo (eg. `sudo phenix ...`) versus indirectly from
+	// a root shell that was itself started via sudo. Only trust `SUDO_USER` if
+	// `SUDO_COMMAND` indicates phenix was the command actually run via sudo.
+	if u.Uid == "0" && sudo != "" && sudoRanPhenix() {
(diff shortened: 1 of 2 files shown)


src/go/cmd/root.go
```diff
@@ -429,9 +429,13 @@ func getCurrentUserInfo() (string, string) {
...
-	// Only trust `SUDO_USER` env variable if we're currently running as root and,
-	// if set, use it to lookup the actual user that ran the sudo command.
-	if u.Uid == "0" && sudo != "" {
+	// `SUDO_USER` (and `SUDO_COMMAND`) remain set in the environment even after
+	// a user has escalated to a root shell (eg. via `sudo su` or `sudo -i`), so
+	// checking `SUDO_USER` alone isn't enough to determine whether phenix was
+	// invoked directly via sudo (eg. `sudo phenix ...`) versus indirectly from
+	// a root shell that was itself started via sudo. Only trust `SUDO_USER` if
+	// `SUDO_COMMAND` indicates phenix was the command actually run via sudo.
+	if u.Uid == "0" && sudo != "" && sudoRanPhenix() {
```
nblair2, line 438: Is it worth pushing all of these checks (the two original `u.Uid == "0" && sudo != ""`) and the documentation down into the called function so its a little cleaner?

Issue #352, opened 2026-08-03
[FEATURE] Additional State of Health checks


GhostofGoes:
### Summary

_Originally posted by @nblair2 in https://github.com/sandialabs/sceptre-phenix/pull/349#pullrequestreview-4834899590_

Spurs some future ideas:
* check if a file **does not** exist
* service checks (assume `systemctl` for linux?)
* docker checks (check health if it has it, otherwise just up)

### Motivation
These would enable more robust validation of experiments without resorting to one-off scripts.

### Proposed Solution

- Implement inverse of `hostFiles` for checking non-existance of files, not sure what this should be called.
- Add `hostServices`. On Windows, use `sc query`, on Linux, `systemctl`.
- Add `dockerContainers`?
- Refactor the SoH app code to iterate over a list of checks, there's a lot of repeated logic that can be consolidated and made more robust ([see this comment](https://github.com/sandialabs/sceptre-phenix/pull/349#discussion_r3696061185)).

### Alternatives Considered
No.

### Additional Context
N/A

### Checklist
- [x] I have included no proprietary/sensitive information in my issue.

Th3BanHamm3r:
@GhostofGoes other than the Docker deployment, I would also look at potentially look at adding SOH to scale deployments. One version of this would look at the SOH of builtin VM deployments. The other would look at the SOH of KVM containers inside a minimeta host. Not sure how you would want to check for services, IPs, etc. (maybe mm cc vm info/exec). Just an idea.

PR #355, opened 2026-08-03
feat(soh): add host file absence checks

GhostofGoes:
## Description
Adds `hostFilesAbsent` config to the `soh` app — the inverse of `hostFiles` — to verify that a file does **NOT** exist on a host, per the first item requested in #352.

- New `hostFilesAbsent: map[hostname][]path` scenario app metadata field
- New `files-absent` check (enabled by default, can be disabled via runtime `checks` selection like other checks)
- New `FilesAbsent` field on `HostState`, included in `AllStates()`
- SoH UI details modal now renders a "Files (Absent)" table alongside the existing "Files" table

## Related Issue
Addresses part of #352 (file non-existence checks). The remaining items in that issue (`hostServices`, `dockerContainers`, and refactoring the repeated SoH check logic) are out of scope for this PR and can be tracked as follow-ups.

## Type of Change
- [x] New feature

## Testing
- `go test ./api/soh/...` (added tests: metadata decode for `hostFilesAbsent`, `HostState.AllStates()` now includes `FilesAbsent`)
- `go build ./api/soh/...`
- `gofmt -l` clean

## Checklist
- [x] I have tested my code.

### DIFF
--- a/src/go/api/soh/util.go
+++ b/src/go/api/soh/util.go
@@ -605,13 +605,29 @@ func (s *SOH) waitForReachabilityTest(ctx context.Context, ns string, checks map
 	return wg.ErrCount > 0
 }
 
-func (s *SOH) waitForFileTest(ctx context.Context, ns string) bool {
+// fileExistenceTestFn matches the signature shared by fileTest and fileAbsentTest.
+type fileExistenceTestFn func(ctx context.Context, wg *mm.StateGroup, ns string, node ifaces.NodeSpec, path string)
+
+// waitForFileExistenceTest runs the given file existence test (fileTest or
+// fileAbsentTest) for each host/path pair in hostFiles, then records results
+// into the host state field selected by assign. This is shared by
+// waitForFileTest and waitForFileAbsentTest since they only differ by which
+// test is run, what gets logged, and where results are stored.
+func (s *SOH) waitForFileExistenceTest(
+	ctx context.Context,
+	ns string,
+	hostFiles map[string][]string,
+	test fileExistenceTestFn,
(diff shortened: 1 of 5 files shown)


src/go/api/soh/util.go
nblair2, line 1491: Interacting with the mm and properly escaping quotes is always a pain. Something I started doing on another project is b64 encode the command on each side. It seems to be working pretty well. The downside is that `mm cc commands` is going to show that base64 version, so you'll only want to do this if you assume the user is looking at commands through your calling program, not raw `mm`.

src/go/api/soh/util.go
```diff
@@ -1449,6 +1514,43 @@ func fileCheckCommand(osType, path string) string {
...
+	path string,
+) {
+	var (
+		host = node.General().Hostname()
+		meta = map[string]any{"host": host, "path": path}
+	)
+
+	retries := 5
+	expected := func(resp string) error {
+		if strings.TrimSpace(resp) == "present" {
```
nblair2, line 1510: This broke my brain a little but makes sense because we are re-using the existing `fileCheckCommand`.

nblair2 (approved):
LGTM.

nblair2:
Lint failing.

GhostofGoes:
> Lint failing.

Sigh. I need to get the prek pre-commit PR finished already. Will address this week.

GhostofGoes:
Fixed lint issues

nblair2 (approved):

cmulk:
@GhostofGoes  sorry now there is a merge conflict after the vue3 update

PR #320, opened 2026-06-09
feat!: upgrade to Vue 3

nblair2:
# feat!: upgrade to Vue 3

> **Original work by @jacdavi and @causand22 (#236).** This PR rebases their Vue 3
> migration onto current `main`, finishes the conversion, and resolves review
> feedback. They are the sole authors of the squashed commit — full credit for the
> migration is theirs. Supersedes #236.

## Description
Updating the front end to use Vue 3 and newer versions of all tools. Quick overview:
- Node version updated to 24
- Removing yarn for build process, using npm
- Vue 2 updated to Vue 3
  - store: Vuex -> Pinia
  - requests: vue resources -> axios
  - build tool: vue-cli -> vite
- buefy updated to v1.0 (vue 3 support)
- Code organization, moved page base components to views/, separated some views into multiple components
- Separate CSS files
- code formatted using prettier
- pages now load dynamically (making initial page loading quicker)
  - Config page has components load dynamically (editor dependencies) to make list load faster.

## Related Issue
- closes #160
- closes #122
- replaces #236 
- docs in https://github.com/sandialabs/sceptre-phenix-docs/pull/54
- additional fixes:
  * https://github.com/sandialabs/sceptre-phenix/pull/356
  * https://github.com/sandialabs/sceptre-phenix/pull/357
  * https://github.com/sandialabs/sceptre-phenix/pull/358
  * https://github.com/sandialabs/sceptre-phenix/pull/359

## Type of Change
- [X] New feature

## Checklist
- [x] This PR conforms to the process detailed in the [Contributing Guide](https://github.com/sandialabs/sceptre-phenix/tree/main/.github/CONTRIBUTING.md).
- [x] I have included no proprietary/sensitive information in my code.
- [x] I have performed a self-review of my code.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have made corresponding changes to the documentation.
- [x] My changes generate no new warnings.
- [x] I have tested my code.

## Additional Notes
This is squashed into 2 commits:
* 8ee8c8ac04d57101780eb889ed54c12469629975 - (squashed) the original work from #236 
* 87caaf27621f32e1c336176f474baac3caa02fd8 - (squashed) changes, regressions, AI review addressed, etc.

PR #356, opened 2026-08-05
fix(ui): minimega console hangs

nblair2:
## Description
The web UI's minimega console has been dead since 85f19b3: the `phenix mm --attach` process is started with the HTTP request's context, which net/http cancels as soon as the `POST /console` response is written

This PR:
- starts the console with `context.WithoutCancel(r.Context())` so it outlives the spawning request
- reaps the process when it exits and removes its dead pty from the map
- fixes a latent deadlock: `ResizeConsole` and `WsConsole` returned with `ptyMu` still locked on the "pty not found" path

## Related Issue
Reported by @GhostofGoes while testing #320 (broken on `main` with the Vue 2 UI as well, hence the standalone fix)

## Type of Change
- [x] Bugfix

## Checklist
- ~I have made corresponding changes to the documentation.~
- [x] I have tested my code.

## Additional Notes
🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/go/web/handlers.go
+++ b/src/go/web/handlers.go
@@ -3795,7 +3795,14 @@ func CreateConsole(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	cmd := exec.CommandContext(r.Context(), phenix, "mm", "--attach") //nolint:gosec // Command injection via taint analysis
+	// the console must outlive this request; r.Context() is canceled as soon
+	// as the response is written, which would SIGKILL the attach process
+	cmd := exec.CommandContext(
+		context.WithoutCancel(r.Context()),
+		phenix,
+		"mm",
+		"--attach",
+	) //nolint:gosec // Command injection via taint analysis
 
 	tty, err := pty.Start(cmd)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


GhostofGoes:
So it was @glattercj's everything bagel at fault?

src/go/web/handlers.go
```diff
@@ -3795,7 +3795,14 @@ func CreateConsole(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	cmd := exec.CommandContext(r.Context(), phenix, "mm", "--attach") //nolint:gosec // Command injection via taint analysis
+	// the console must outlive this request; r.Context() is canceled as soon
+	// as the response is written, which would SIGKILL the attach process
+	cmd := exec.CommandContext(
```
GhostofGoes, line 3800: Would this session overlap with a session launched from a terminal? Maybe spawn a separate mm socket?

GhostofGoes: Nevermind this, it appears to launch a new session.

GhostofGoes:
This is tested working for me. I'm also able to reliably use from the UI and from the CLI simultaneously. Note that when the console tab is closed, it resets/clears the session, so any user commands run are lost. This is fine, but could be confusing behavior for some users. Future work would be to make a single session persistent for given user/browser, and automatically reap when last connection from source token/user is closed.

PR #329, opened 2026-07-18
feat(tap): add stage option to run tap app at pre-start

cmulk:
# feat(tap): add stage option to run tap app at pre-start

## Description
The `tap` app previously only created its host taps during the **post-start** stage. This adds a `stage` metadata option so taps can instead be created during the **pre-start** stage.

- New `stage` field on the tap app metadata. Valid values are `pre-start` and `post-start`.
- Defaults to `post-start` when unset, so existing scenarios are unaffected.
- The tap-creation logic was refactored out of `PostStart` into a shared `createTaps` helper that both `PreStart` and `PostStart` call; it only runs during the configured stage.

## Related Issue
If applicable, please link to the issue here (e.g., #123).

## Type of Change
Please select the type of change your pull request introduces:
- [ ] Bugfix
- [x] New feature
- [ ] Documentation update
- [ ] Other (please describe):

## Checklist
- [x] This PR conforms to the process detailed in the [Contributing Guide](https://github.com/sandialabs/sceptre-phenix/tree/main/.github/CONTRIBUTING.md).  
- [x] I have included no proprietary/sensitive information in my code. 
- [x] I have performed a self-review of my code.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have made corresponding changes to the documentation.
- [x] My changes generate no new warnings.
- [x] I have tested my code.

## Additional Notes
Please provide any additional information or context for your pull request here.

PR #365, opened 2026-08-12
fix(experiment): avoid cleanup race when stopping during delayed VM start

cmulk:
# Fix cleanup race when stopping an experiment during delayed VM start

## Description
When an experiment configured with delayed VMs (time-delay or C2-gated
start, see #329) is stopped while post-start apps are still waiting for
those delayed VMs to start, the context.Context passed to
handleDelayedVMs gets canceled by the stop request. Previously this
cancellation was treated the same as any other delayed-VM failure:

- handleDelayedVMs called mm.ClearNamespace itself, and
- the Start goroutine called Stop(...) a second time

both of which raced with the Stop call already in progress (triggered
by the user's stop request), which is independently running app
Cleanup stages (e.g. the tap app deleting the taps it created in
PreStart) and then clearing the namespace. The result is a race where
the namespace/taps can be torn down out from under the in-progress
Cleanup stage, surfacing as spurious errors about tap/VM resources
that no longer exist.

This PR makes handleDelayedVMs and the Start goroutine recognize
context cancellation as "a stop is already in flight elsewhere" rather
than a fresh failure to react to, so they no longer race with the
already-owned cleanup:
- handleDelayedVMs skips its own mm.ClearNamespace call when the
  aggregated error is due to ctx being canceled.
- The Start goroutine skips its redundant Stop(...) call when
  handleDelayedVMs failed because ctx was canceled.

A unit test (handle_delayed_vms_test.go) was added that reproduces
the race directly: with a canceled context and a pending delayed VM,
it asserts mm.ClearNamespace is not invoked. It fails against the
pre-fix code and passes with the fix, including under -race.

### Update: tap app cleanup failure when post-start never ran
A related issue surfaced once the race above was fixed: if the tap app
is configured with `stage: post-start` (the default) and the
experiment is stopped before delayed VMs finish starting, `PostStart`
(and thus `createTaps`) never executes, so no status is ever recorded
for the tap app. The subsequent `Cleanup` call would unconditionally
call `ParseAppStatus`, which failed with:

```
cleaning up app experiments: applying user app tap for action cleanup: getting experiment status for tap app: missing status for app tap
```

`Cleanup` now checks `exp.Status.AppStatus()` for an entry before
parsing it, and treats a missing entry as a no-op, since there are no
taps to delete in that case.

A test (tap_test.go) reproduces this exact error message against the
pre-fix code and passes with the fix, plus a companion test asserting
`Cleanup` still processes a legitimate (even if empty) recorded status.

## Related Issue
Follow-up to #329.

## Type of Change
- [x] Bugfix
- [ ] New feature
- [ ] Documentation update

## Checklist
- [x] This PR conforms to the process detailed in the Contributing Guide.
- [x] I have tested my code.

## Additional Notes
- Added src/go/api/experiment/handle_delayed_vms_test.go, following the
  existing test-double convention in src/go/app/user_test.go (swapping
  mm.DefaultMM with a struct embedding the mm.MM interface).
- Added src/go/app/tap_test.go covering the tap app Cleanup no-op fix.
- Verified via go build/go vet/go test -race for phenix/api/experiment
  and phenix/app (the only packages touched).
- Confirmed both new tests fail on the pre-fix code and pass with the fix.
- Commits squashed into a single commit for review.

Manually tested multiple combos of tap app with delayed start VMs and premature experiment stops. Confirmed that stops and cleanup occurs without error.

### DIFF
--- a/src/go/api/experiment/experiment.go
+++ b/src/go/api/experiment/experiment.go
@@ -816,6 +816,14 @@ func Start(ctx context.Context, opts ...StartOption) error {
 				if err != nil {
 					o.errChan <- fmt.Errorf("handling delayed VMs: %w", err)
 
+					// If the context was canceled, the experiment is already being
+					// stopped elsewhere (e.g. via the web UI, which cancels this
+					// context before calling Stop). Calling Stop again here would
+					// race with that in-progress stop and its cleanup, so skip it.
+					if errors.Is(ctx.Err(), context.Canceled) {
+						return
+					}
+
 					err = Stop(exp.Spec.ExperimentName())
 					if err != nil {
 						o.errChan <- fmt.Errorf("stopping experiment: %w", err)
@@ -1245,6 +1253,16 @@ func handleDelayedVMs(
 	}
 
 	if errs != nil {
+		// If the context was canceled, the experiment is already being (or is
+		// about to be) stopped elsewhere (e.g. a user requested the experiment
+		// be stopped while delayed VMs were still pending). That in-progress
+		// stop already owns tearing down the namespace, so clearing it here too
+		// would race with it and can delete resources (such as taps created by
+		// the tap app) out from under that cleanup.
+		if errors.Is(ctx.Err(), context.Canceled) {
+			return errs //nolint:wrapcheck // returning multierror
+		}
+
 		err2 := mm.ClearNamespace(ns)
 		if err2 != nil {
 			errs = multierror.Append(errs, fmt.Errorf("killing experiment VMs: %w", err2))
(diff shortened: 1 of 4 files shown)


GhostofGoes:
I'm able to reproduce this issue. Will test fix if I get a chance.

GhostofGoes:
tested fix seems to resolve for tap

mgaliar (approved):
LGTM!

GhostofGoes:
Send it!

activeshadow:
You all have been killing it with all these great fixes lately! ❤️

GhostofGoes:
@activeshadow sorry you've been assigned to every PR. If you want, I'm happy to update the CODEOWNERS file.

activeshadow:
@GhostofGoes I'm fine with it. I'm actively working on a way to get back to doing this kinda work anyway so it's all good.

PR #366, opened 2026-08-12
fix(disk): populate experiment field for images in use

### LINKED ISSUE #344: [BUG] UI: Experiment field never populated for disks
### Description

In the Disks view in the UI, the Experiment field always shows "N/A", even if the image is in active use by a running experiment.

### Steps to Reproduce
1. Create and start an experiment
2. Navigate to Disks page
3. Open a disk that's in use, observe the "Experiment" field

### Expected Behavior
It should show the name of the experiment using the image, or a list of experiments if multiple experiments are using it. This should also be the case for stopped experiments, IMO.

### Actual Behavior
Just shows "N/A"

### Screenshots

Image

Image

### Environment
- **Operating System**: Ubuntu 22
- **Browser**: Chrome
- **Version**: Hmm

### Additional Context
Probably would be faster to just fix than open this bug report but here we are.

### Checklist
- [x] I have included no proprietary/sensitive information in my issue.

stark256-spec:
Closes #344.

## Problem

In the Disks view, an image's **Experiment** field always shows `N/A`, even when the image is in active use by a running experiment.

## Root cause

`disk.GetImages` builds its result in two phases:

1. `getAllFiles` adds every image from the minimega `file list` with **no** experiment association (`Experiment` left `nil`).
2. `getTopologyFiles` then walks each experiment's topology — but it only *adds* images that are **not already** in the map:

   ```go
   if _, ok := details[filepath.Base(path)]; !ok {
       for _, image := range resolveImage(path) { ... }
   }
   ```

Because phase 1 has usually already inserted the image, phase 2 skips it and never records the experiment, so `Experiment` stays `nil` and the UI renders `N/A` (`Disks.vue`: `{{ disk.experiment || "N/A" }}`).

## Fix

`getTopologyFiles` now records the referencing experiment on each topology image via a new `markExperimentImages` helper, which **updates entries already present** in the map rather than skipping them. The `Experiment *string` type and JSON shape are unchanged, so no frontend change is needed.

When `GetImages` is called without an experiment name it walks every experiment; if the same image is referenced by more than one, the last one processed wins (single `*string`). That already fixes the reported bug (a real experiment name instead of `N/A`); a `[]string` for the multi-experiment case would be a larger, API-breaking change and is left as a follow-up.

## Tests

Added `src/go/api/disk/disk_test.go` (the package previously had none) covering the helper:

- `TestMarkExperimentImages` — a referenced image gets its `Experiment` set; unreferenced images are untouched; paths not already in the map are not added.
- `TestMarkExperimentImagesUpdatesExisting` — an existing (stale) association is updated.

```
$ go test ./api/disk/
ok  	phenix/api/disk
```

`gofmt` and `go vet ./api/disk/` are clean.

### DIFF
--- a/src/go/api/disk/disk.go
+++ b/src/go/api/disk/disk.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"path/filepath"
 	"regexp"
+	"sort"
 	"strconv"
 	"strings"
 
@@ -90,12 +91,16 @@ func (MMDiskFiles) GetImages(expName string) ([]Details, error) {
 	// Using a map here to weed out duplicates.
 	details := make(map[string]Details)
 
+	// Tracks the set of experiments whose topology references each image (keyed
+	// by image name), so an image used by multiple experiments can list them all.
+	expsByImage := make(map[string]map[string]struct{})
+
 	// Add all the files from the minimega files directory
 	getAllFiles(details)
 
 	// Add all files defined in the experiment topology if given; otherwise check all experiments
 	if len(expName) > 0 {
-		err := getTopologyFiles(expName, details)
+		err := getTopologyFiles(expName, details, expsByImage)
 		if err != nil {
 			return nil, err
 		}
@@ -106,12 +111,16 @@ func (MMDiskFiles) GetImages(expName string) ([]Details, error) {
 		}
 
 		for _, exp := range experiments {
-			err = getTopologyFiles(exp.Metadata.Name, details)
+			err = getTopologyFiles(exp.Metadata.Name, details, expsByImage)
 			if err != nil {
 				return nil, err
 			}
 		}
 	}
+
+	// Record the referencing experiment(s) on each image, sorted and joined so
+	// images used by more than one experiment list all of them deterministically.
+	setExperimentImages(details, expsByImage)
 
 	var images []Details
 	for name := range details {
@@ -152,7 +161,7 @@ func getAllFiles(details map[string]Details) {
 }
 
 // Retrieves all the unique image names defined in the topology.
-func getTopologyFiles(expName string, details map[string]Details) error {
+func getTopologyFiles(expName string, details map[string]Details, expsByImage map[string]map[string]struct{}) error {
 	// Retrieve the experiment
 	exp, err := experiment.Get(expName)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


GhostofGoes:
Can you show screenshots demonstrating the fix works in the UI?

stark256-spec:
image-2
image-1
changes
Here you go — captured against a real minimega + phenix stack (the official `phenix:main` + `minimega:master` containers on a Linux host), not a mock.

Setup: created a disk image `ubuntu-server.qcow2` and an experiment `power-grid-demo` whose topology references it (experiment *created, not started* — matching the "should also work for stopped experiments" note in the issue).

- **Before** (stock `main`, commit `1a4db47`): the Disks detail modal shows **Experiment: N/A**.
- **After** (this PR, commit `388a24d`): same disk, same experiment — now shows **Experiment: power-grid-demo**.



`GET /api/v1/disks` confirms it at the API level too: `"experiment": null` before → `"experiment": "power-grid-demo"` after.

nblair2:
Thanks for fixing! What happens if multiple experiments reference the same disk?

stark256-spec:
Good catch — currently the `experiment` field is a single string, so if several experiments reference the same disk only one name shows (whichever `experiment.List()` yields last), which isn't deterministic. I just reproduced that on the stack above with a second experiment on the same disk.

I'll update the PR so it lists **all** referencing experiments — collecting them and joining sorted (e.g. `power-grid-demo, substation-demo`) in the existing string field, which keeps the API and the `Disks.vue` binding unchanged. (If you'd rather have a structured `experiments: []string`, I'm happy to do that instead and update the `|| "N/A"` fallback in the modal.)

stark256-spec:
Implemented the multi-experiment handling in e123daf. An image referenced by several experiments now lists **all** of them, sorted and comma-joined, in the same `experiment` string field — so it's deterministic and no longer depends on `experiment.List()` order. The `Details.Experiment` type and JSON shape are unchanged, so no frontend change is needed.

Captured live on the same stack, with both `power-grid-demo` and `substation-demo` referencing `ubuntu-server.qcow2`:



Added `TestSetExperimentImagesListsAllSorted` covering the sorted multi-experiment case.

GhostofGoes:
@stark256-spec can you share screenshot of the multi-experiment case? Also can u DM me on Teams/Email? Would like to chat more 😃 .

stark256-spec:
I don't have your email since it's public comment,  if you're open to sharing, a burner/ghost email works too, no need for your personal one! || Killed the stack I thought posted image, bringing it up .. it takes a moment on my end ..

stark256-spec:
Here's the multi-experiment case, captured live on the same phenix + minimega stack. Both `power-grid-demo` and `substation-demo` reference `ubuntu-server.qcow2`, and the Disks modal now shows **Experiment: power-grid-demo, substation-demo** (sorted, comma-joined):

image

`GET /api/v1/disks` confirms it: `"experiment": "power-grid-demo, substation-demo"`.

GhostofGoes (approved):

GhostofGoes:
> I don't have your email since it's public comment, if you're open to sharing, a burner/ghost email works too, no need for your personal one! || Killed the stack I thought posted image, bringing it up .. it takes a moment on my end ..

It should be in `git log`

stark256-spec:
emailed you on nlr domain.gov, probably went to spam or never made it.

cmulk:
looks like there are conflicts now unfortunately, can you rebase/fix conflicts and squash commits please

stark256-spec:
Done — rebased onto latest `main` and squashed to a single commit (`0a12af9`). The only conflict was the import block, from the new `regexp`/`ResizeDisk` size validation landing alongside my `sort` import; resolved so both are kept. `go build`, `go vet`, `gofmt`, and the `api/disk` tests all pass. Should be good to go now.

PR #407, opened 2026-09-04
feat(tunneler): add local web interface

activeshadow:
# Pull Request Title

## Description
**Tunneler Web Interface**: Add a local web dashboard for listing, enabling, disabling, and moving listeners, with live WebSocket updates.

In addition, this PR also includes the following fixes for Tunneler:
- **Listener State Synchronization**: Centralized local listener tracking behind a synchronized manager to prevent concurrent map and state access.
- **Operation Reporting**: Return errors for unknown listeners and local port conflicts instead of reporting successful operations.
- **Argument Handling**: Reject malformed command arguments and Unix socket payloads without panicking.
- **Listener Moves**: Restore an active listener's original port when moving to a new port fails.
- **Listener IDs**: Maintain a direct ID index and prevent IDs from being reused during a tunneler session.
- **Listener Listing**: Return local listeners in deterministic ID order.
- **Duplicate Listeners**: Treat repeated listener creation events as idempotent without replacing active listeners.
- **Unix Socket Protocol**: Replace Gob messages with JSON and typed listener action payloads.

## Related Issue
N/A

## Type of Change
- [ ] Bugfix
- [x] New feature
- [ ] Documentation update

## Checklist
- [x] I have tested my code.

## Additional Notes
N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,6 +144,10 @@ make install-dev # Install development and build dependencies
 make clean       # Clean build artifacts
 ```
 
+The `phenix-tunneler serve` command can provide a local web interface for
+listing and managing listeners. It is disabled by default; enable it with
+`--web-listen 127.0.0.1:8080`. The address can be changed with that option.
+
 ### Build
 
 To build the phēnix core services locally:
--- a/src/go/tunneler/client.go
+++ b/src/go/tunneler/client.go
@@ -2,7 +2,7 @@ package main
 
 import (
 	"context"
-	"encoding/gob"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"math/rand/v2"
@@ -14,8 +14,8 @@ import (
 type client struct {
 	conn net.Conn
 
-	enc *gob.Encoder
-	dec *gob.Decoder
+	enc *json.Encoder
+	dec *json.Decoder
 }
 
 func newClient() (*client, error) {
@@ -33,8 +33,8 @@ func newClient() (*client, error) {
 		return nil, fmt.Errorf("dialing phenix unix socket %s: %w", sockPath, err)
 	}
 
-	cli.enc = gob.NewEncoder(cli.conn)
-	cli.dec = gob.NewDecoder(cli.conn)
+	cli.enc = json.NewEncoder(cli.conn)
+	cli.dec = json.NewDecoder(cli.conn)
 
 	return cli, nil
 }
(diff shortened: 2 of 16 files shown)


README.md
```diff
@@ -144,6 +144,10 @@ make install-dev # Install development and build dependencies
 make clean       # Clean build artifacts
 ```
 
+The `phenix-tunneler serve` command can provide a local web interface for
```
GhostofGoes, line 147: This belongs in the docs? Fine with it in readme too.

src/go/tunneler/client.go
```diff
@@ -118,17 +103,13 @@ func (c client) deactivateLocalListener(id int) error {
 	msg := Message{ //nolint:exhaustruct // partial initialization
 		MID:     int(rand.Uint64()), //nolint:gosec // weak random number generator
 		Type:    DEACTIVATE,
-		Payload: id,
+		Payload: marshalPayload(
```
GhostofGoes, line 106: Why can't json.Marshall be used here? Sorry if it's a dumb Go question.

src/go/tunneler/main.go
```diff
@@ -232,6 +228,14 @@ var serveCmd = &cobra.Command{ //nolint:gochecknoglobals // cobra command
 			return fmt.Errorf("starting unix socket: %w", err)
 		}
 
+		if webListen != "" {
+			if err := startWebServer(webListen); err != nil {
+				return fmt.Errorf("starting web server: %w", err)
+			}
+
+			fmt.Printf("tunneler web interface available at http://%s\n", webListen) //nolint:forbidigo // CLI output
```
GhostofGoes, line 236: No https?

GhostofGoes (changes requested):

src/go/tunneler/client.go
activeshadow: Because `json.Marshal` returns a value and an error (so it can't be used inline). `marshalPayload` is just a helper to ignore any marshaling errors (there shouldn't be any since the payload being marshaled is always generated by code, not by users).

src/go/tunneler/main.go
activeshadow: No, the phenix-tunneler app only runs locally on a user's machine.

GhostofGoes (approved):

README.md
cmulk: my preference would be to add it to the Tunneler tab in phenix

cmulk (changes requested):
One requested change below about documenting the new function in the Tunneler tab. Otherwise I tested on my end and looks good

PR #420, opened 2026-09-17
fix(startup): restore Windows startup wrapper and scheduler injects

cmulk:
## Description

Commit `3cdd399` added minimega C2 delivery for generated startup scripts, but in the process it also removed the Windows `phenix-startup.ps1` wrapper injection and the Start Menu `startup_scheduler.cmd` injection. Windows nodes that receive their startup scripts by disk injection depend on both: the wrapper iterates and runs every script staged under `/phenix/startup`, and the Start Menu scheduler registers that wrapper as an on-start scheduled task. This requires the startup scheduled task to be pre-configured in the base image (which I know is better but we still depend on these injections in a lot of cases). Without them, disk-injected Windows nodes silently stopped running their startup scripts (hostname, timezone, interface, and domain configuration).

This PR restores both injections as the default behavior for Windows nodes, while leaving the C2 delivery path introduced by `3cdd399` intact.

Relevant changes:

- Added `addWindowsStartupWrapperInjects` in `src/go/app/startup.go`, which restores the `phenix-startup.ps1` and `startup-scheduler.cmd` embedded assets and injects them through the existing `addStartupInject` helper so they honor the same suppression rules and `0644` permissions as every other startup-owned injection.
- Renamed `legacyWindowsStartupWrapperDst` to `windowsStartupWrapperDst` and added `windowsStartupWrapperAsset` / `windowsSchedulerAsset` constants, since these destinations are no longer legacy.
- Renamed `removeLegacyWindowsStartupInjections` to `removeWindowsStartupWrapperInjections` and dropped the associated migration `TODO`. It is now the deliberate cleanup path for C2-delivered nodes rather than a temporary shim.
- Added the wrapper and scheduler destinations to `removeStartupInjections` so suppression and repeated `pre-start` runs stay idempotent and never leave stale entries behind.
- When C2 startup delivery is in use — either the `phenix/startup-via-cc` annotation or a first drive with `inject_partition: 0` — both injections are skipped and any previously injected entries are removed, because C2 executes the generated script directly and the wrapper would be redundant.

No new annotations were introduced. The wrapper and scheduler injections are unconditional for Windows nodes that are not using C2 delivery, which matches the behavior prior to `3cdd399`.

## Related Issues/PRs

Regression introduced by `3cdd399`.

## Type of Change

- [x] Bugfix (`fix`)
- [ ] Feature (`feat`)
- [ ] Documentation (`docs`)
- [ ] Refactor (`refactor`)
- [ ] Chore (CI, build, dependencies, etc.) (`chore`)

## Checklist

- [x] I have included no proprietary/sensitive information in my code or the PR.
- [x] I have tested my code (describe below).

## Testing

Tested on a simple experiment with a windows VM, confirmed startup cmd and wrapper have been restored as injects, and confirmed they are not injected when using C2 startup

Unit tests were added and updated in `src/go/app/startup_test.go`:

- `TestStartupPreStartWindowsWrapperInjections` — table-driven over the two delivery modes, asserting that a disk-injected Windows node receives both the wrapper and Start Menu scheduler injections, and that a node annotated with `phenix/startup-via-cc` receives neither.
- `TestStartupPreStartRemovesWindowsWrapperInjectionsForC2Delivery` — asserts that pre-existing wrapper and scheduler injections are removed when C2 delivery is selected, covering the case where a node is switched from disk injection to C2 between runs.
- `TestStartupPreStartWindowsC2Commands` — updated to account for the restored injections while still asserting the `send` and `exec-once` C2 commands from `3cdd399` are unchanged.
- `TestStartupPreStartAnnotationPreservesUnrelatedInjections` — updated to confirm the broader removal set still only touches startup-owned destinations and leaves unrelated user injections alone.

## Additional Notes

`addWindowsStartupWrapperInjects` intentionally restores both template assets to the experiment's `startup/` directory even when the injections are suppressed. This mirrors how the Linux startup scripts are handled, where script generation and the injection decision are kept separate. The consequence is that a `phenix/startup-via-cc` node still has the two files written to disk without them being injected. An early return on the suppress flag would avoid those writes; it was omitted for consistency, and can be added if reviewers prefer.

Documentation was updated in `skills/phenix/SKILL.md`. The stale `windows-version` annotation entry was replaced with documentation for `phenix/startup-via-cc`, describing both the annotation-driven and `inject_partition: 0` paths into C2 delivery.


mgaliar (approved):

cmulk:
@activeshadow I agree and prefer the startup task and wrapper to be pre-configured in the base images, but we still have lots of vms in our environment that depend on these injections

cmulk:
on a related note,  https://github.com/sandia-minimega/minimega/pull/1667 should fix the startup exec-once never getting run (at least I'm seeing that consistently on windows when trying to use the C2 startup mode)

